PR #31, opened 2019-12-15
Build libvips with a GitHub Actions workflow

joschi:
The version of libvips can be defined in the `LIBVIPS_VERSION` file in the root of the repository.

Every time a new tag starting with 'v*', such as `v8.8.1` is being pushed, the "build-release" workflow will be triggered and do the following things:

* Create a release for the tag on GitHub
* Build libvips with the version provided in `LIBVIPS_VERSION` for every platform
* Upload the release artifacts to the GitHub release which was created before


joschi:
Example run: https://github.com/joschi/sharp-libvips/commit/457779bde13a81994530684b53eb1973a1d731a1/checks
Example release: https://github.com/joschi/sharp-libvips/releases/tag/refs%2Fpull%2F1%2Fmerge

.github/workflows/build-release.yml
```diff
@@ -0,0 +1,63 @@
...
+    - name: Create Release
+      id: create-release
+      uses: actions/create-release@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        tag_name: ${{ github.ref }}
+        release_name: Release ${{ github.ref }}
+        draft: false
+        prerelease: false
```
lovell: I usually label everything as `pre-release` initially to ensure there's some kind of "do not use this in production" warning. This allows some time to test the tarballs on various platforms before manually marking it as production-ready and publishing new versions of sharp that depend upon it.

The existing build script should probably do more verification of the resultant tarballs, which would then provide more confidence in their production readiness, but we're not quite there yet.

joschi: I've changed the attribute to mark the GitHub Release as pre-release.

With regards to testing the tar-balls, we could create that as an additional job after the build and pass the artifacts between jobs.
Do you have a suggestion how to verify the tar-balls?

lovell: Basic verification might be as simple as ensuring `ldd-cpp.so` has zero "not found" dependencies, at least for Linux. A more complex process could checkout sharp, force it to use dependencies from a given tarball and run all its unit tests. Let's merge these improvements first as they are independent of any future verification step.

PR #33, opened 2019-12-29
Update licences

kleisauke:
This PR adds the following licences that were missing:
- gdk-pixbuf - LGPLv2.1;
- orc - [orc License](https://gitlab.freedesktop.org/gstreamer/orc/blob/master/COPYING) (BSD-like) 
   (could also be changed to this SPDX identifier: `BSD-2-Clause AND BSD-3-Clause`, [see discussion](https://github.com/conda-forge/gstreamer-orc-feedstock/pull/4/files/96bfda04d089eb224da66b1ab63179a5c79e6f50#r353797464))

And removes the license for libuuid, as it is no longer required by fontconfig. I've also updated a couple of URLs to a more secure variant.

For easier reviewing, I've made the column width change in a separate commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,9 +50,10 @@ Use of libraries under the terms of the LGPLv3 is via the
 |---------------|----------------------------------------------------------------------------------------------------------|
 | cairo         | Mozilla Public License 2.0                                                                               |
 | expat         | MIT Licence                                                                                              |
-| fontconfig    | [fontconfig Licence](https://cgit.freedesktop.org/fontconfig/tree/COPYING) (BSD-like)                    |
-| freetype      | [freetype Licence](http://git.savannah.gnu.org/cgit/freetype/freetype2.git/tree/docs/FTL.TXT) (BSD-like) |
+| fontconfig    | [fontconfig Licence](https://gitlab.freedesktop.org/fontconfig/fontconfig/blob/master/COPYING) (BSD-like)|
+| freetype      | [freetype Licence](https://git.savannah.gnu.org/cgit/freetype/freetype2.git/tree/docs/FTL.TXT) (BSD-like)|
 | fribidi       | LGPLv3                                                                                                   |
+| gdk-pixbuf    | LGPLv3                                                                                                   |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,30 +46,31 @@ all of which are compatible with the Apache 2.0 Licence.
...
+| freetype      | [freetype Licence](https://git.savannah.gnu.org/cgit/freetype/freetype2.git/tree/docs/FTL.TXT) (BSD-like) |
+| fribidi       | LGPLv3                                                                                                    |
+| gdk-pixbuf    | LGPLv3                                                                                                    |
+| gettext       | LGPLv3                                                                                                    |
+| giflib        | MIT Licence                                                                                               |
+| glib          | LGPLv3                                                                                                    |
+| harfbuzz      | MIT Licence                                                                                               |
+| lcms          | MIT Licence                                                                                               |
+| libcroco      | LGPLv3 (Windows only)                                                                                     |
```
kleisauke: libcroco also still seems to be used on macOS, see:
https://github.com/Homebrew/homebrew-core/blob/master/Formula/librsvg.rb#L19

Shall I change it into?:
```suggestion
| libcroco      | LGPLv3 (Windows / macOS only)                                                                                     |
```

lovell: Yes please.

lovell:
Thanks for these Kleis.

Perhaps we should split this table out into a separate `LICENSING.md` file that can also be included in the resultant tarballs?

README.md
kleisauke: Resolved with: https://github.com/lovell/sharp-libvips/pull/33/commits/1a37a8b7b43b0ddfb12e2c8a0ff6cb13bef291e7.

kleisauke:
Good idea! Perhaps the file should be named `Third-Party-Notices.md` or `THIRD-PARTY-NOTICES.md`? It seems more common.

lovell:
Thanks, I've split these out in commit df81c75 for future possible inclusion in the tarballs.

PR #44, opened 2020-05-25
Build libvips with its dependencies statically-linked (#39)

### LINKED ISSUE #39: Experiment: provide libvips as a single shared library with its dependencies statically-linked
It looks like macOS doesn't expand `@loader_path` to the actual filesystem path before checking if the rpath of a dylib file has already been loaded. This means that different versions of sharp at different filesystem locations loaded in the same process can conflict when an earlier version is loaded before a later version.

To attempt to deal with this, we can experiment with unpacking prebuilt libvips binaries into versioned directory under `vendor`, e.g. `node_modules/sharp/vendor/8.9.0/lib/...`, which should then ensure the rpath includes the version number.

kleisauke:
This PR adds supports for a single shared libvips library with its dependencies statically-linked as described in issue #39. It also ensures that the packaging script is compatible with macOS.

The `MACOSX_DEPLOYMENT_TARGET` is set to 10.9 (Mavericks). This version can also be upgraded to 10.13 (High Sierra, just like Node 14), since the previous [Homebrew-based packaging script](https://github.com/lovell/package-libvips-darwin) uses this as the minimum deployment target, see:
https://github.com/Homebrew/brew/blob/b9b51421f4c74f27cff38fb722b56dd6e6e74611/Library/Homebrew/extend/os/mac/extend/ENV/std.rb#L78-L79
https://github.com/lovell/package-libvips-darwin/blob/5816c78be31a9311b7a6bab5ac6bb10c846637ee/.travis.yml#L4

Resolves #39 after merging.


README.md
```diff
@@ -35,12 +35,14 @@ The output of libvips' [build-win64-mxe](https://github.com/libvips/build-win64-
 
 ### macOS
 
-See [package-libvips-darwin](https://github.com/lovell/package-libvips-darwin).
```
kleisauke, line 38: We may have to add a depreciation note to the [`package-libvips-darwin`](https://github.com/lovell/package-libvips-darwin) repository after this.

build/lin.sh
```diff
@@ -340,10 +416,12 @@ printf "\"${PLATFORM}\"" >platform.json
 # Add third-party notices
 curl -Os https://raw.githubusercontent.com/lovell/sharp-libvips/master/THIRD-PARTY-NOTICES.md
 
-# Create .tar.gz
-tar czf /packaging/libvips-${VERSION_VIPS}-${PLATFORM}.tar.gz \
+# Create the tarball
+rm -rf lib
+mv lib-filterd lib
+tar chzf ${PACKAGE}/libvips-${VERSION_VIPS}-${PLATFORM}.tar.gz \
```
kleisauke, line 422: The `tar czf` -> `tar chzf` change dereferences symbolic links during packing, which could be useful for https://github.com/lovell/sharp/issues/2197.

build/lin.sh
```diff
@@ -294,16 +332,54 @@ cd ${DEPS}/vips
...
+# https://docs.fedoraproject.org/en-US/packaging-guidelines/#_removing_rpath
+sed -i'.bak' 's|^hardcode_libdir_flag_spec=.*|hardcode_libdir_flag_spec=""|g' libtool
 make install-strip
 
 # Cleanup
 rm -rf ${TARGET}/lib/{pkgconfig,.libs,*.la,cmake}
-cd ${TARGET}/include
 
-# Set RPATH to $ORIGIN
-find ${TARGET}/lib -type f -name "*.so*" -exec sh -c "patchelf --set-rpath '\$ORIGIN' --force-rpath {}" \;
```
kleisauke, line 306: Note that the `patchelf` trick isn't necessary anymore. It's now handled in the `LDFLAGS` above.

build/lin.sh
```diff
@@ -255,36 +287,42 @@ make install-strip
...
 
 mkdir ${DEPS}/svg
 curl -Lks https://download.gnome.org/sources/librsvg/$(without_patch $VERSION_SVG)/librsvg-${VERSION_SVG}.tar.xz | tar xJC ${DEPS}/svg --strip-components=1
 cd ${DEPS}/svg
-sed -i "s/^\(Requires:.*\)/\1 cairo-gobject pangocairo/" librsvg.pc.in
+sed -i'.bak' "s/^\(Requires:.*\)/\1 cairo-gobject pangocairo/" librsvg.pc.in
+# Do not include debugging symbols
+sed -i'.bak' "/debug =/ s/= .*/= false/" Cargo.toml
 # Optimise Rust code for binary size
-sed -i "s/debug = true/debug = false\nopt-level = 's'\nlto = true\ncodegen-units = 1\nincremental = false\npanic = 'abort'/" Cargo.toml
```
kleisauke, line 277: macOS doesn't like new lines (`\n`) replacements within `sed`, see:
https://stackoverflow.com/questions/10748453/replace-comma-with-newline-in-sed-on-macos

.github/workflows/build-release.yml
```diff
@@ -50,10 +58,22 @@ jobs:
...
+    - name: Install dependencies (macOS)
+      run: |
+        pip3 install meson ninja
+        brew install advancecomp automake nasm pkg-config
+      if: matrix.os == 'macos-latest'
+    - name: Setup toolchain (macOS)
+      shell: bash
+      run: |
+        echo "::set-env name=CC::clang"
+        echo "::set-env name=CXX::clang++"
```
kleisauke, line 70: I also tried the Clang from LLVM with:
```yml
        echo "::set-env name=CC::/usr/local/opt/llvm/bin/clang"
        echo "::set-env name=CXX::/usr/local/opt/llvm/bin/clang++"
```
and GCC 9 with:
```yml
        echo "::set-env name=CC::gcc-9"
        echo "::set-env name=CXX::g++-9"
```

But this didn't succeed.

lovell:
Marvellous, thank you Kleis, I'll make time later in the week to review this properly.

build/lin.sh
```diff
@@ -190,101 +231,106 @@ mkdir ${DEPS}/gdkpixbuf
...
-# Clear executable bit from librsvg shared library for WSL support
-execstack -c ${TARGET}/lib/librsvg-2.so || true
+sed -i'.bak' "s/^\(Requires:.*\)/\1 cairo-gobject pangocairo/" librsvg.pc.in
+# Do not include debugging symbols
+sed -i'.bak' "/debug =/ s/= .*/= false/" Cargo.toml
+# LTO optimization does not work for staticlib+rlib compilation
+sed -i'.bak' "s/, \"rlib\"//" librsvg/Cargo.toml
+./configure --host=${CHOST} --prefix=${TARGET} --enable-static --disable-shared --disable-dependency-tracking \
+  --disable-introspection --disable-tools --disable-pixbuf-loader ${DARWIN:+--disable-Bsymbolic}
+make install-strip RUST_TARGET_SUBDIR="${RUST_TARGET}/release"
```
lovell, line 328: Was the intent here to declare an env var for the command?
```diff
- make install-strip RUST_TARGET_SUBDIR="${RUST_TARGET}/release" 
+ RUST_TARGET_SUBDIR="${RUST_TARGET}/release" make install-strip
```

build/lin.sh
```diff
@@ -293,17 +339,54 @@ cd ${DEPS}/vips
...
+sed -i'.bak' 's|^hardcode_libdir_flag_spec=.*|hardcode_libdir_flag_spec=""|g' libtool
 make install-strip
 
 # Cleanup
 rm -rf ${TARGET}/lib/{pkgconfig,.libs,*.la,cmake}
-cd ${TARGET}/include
 
-# Set RPATH to $ORIGIN
-find ${TARGET}/lib -type f -name "*.so*" -exec sh -c "patchelf --set-rpath '\$ORIGIN' --force-rpath {}" \;
+mkdir ${TARGET}/lib-filterd
```
lovell, line 350: Nitpick: `filterd` sounds like a subgenre of French house music - did you mean `filtered`?

build.sh
```diff
@@ -18,14 +18,33 @@ if [ $# -lt 1 ]; then
...
   echo
   exit 1
 fi
 VERSION_VIPS="$1"
 PLATFORM="${2:-all}"
 
+# macOS
+# Note: we intentionally don't build these binaries inside a Docker container
+if [ $PLATFORM = "darwin-x64" ]; then
+  export VERSION_VIPS
```
lovell, line 31: It might be worth adding a check that `uname -s` is "Darwin" here in case someone tries to run this on Linux.

build.sh
```diff
@@ -18,14 +18,33 @@ if [ $# -lt 1 ]; then
...
+  export PLATFORM
+  export RUST_TARGET="x86_64-apple-darwin"
+
+  # Added -fno-stack-check to workaround a stack misalignment bug on macOS 10.15
+  # See:
+  # https://forums.developer.apple.com/thread/121887
+  # https://trac.ffmpeg.org/ticket/8073#comment:12
```
lovell, line 38: Does this mean that AVX is enabled for the macOS build? If so, do we know how? We may need to ensure more recent intrinsics such as AVX512 are not enabled to ensure support for older hardware.

.github/workflows/build-release.yml
```diff
@@ -39,6 +42,11 @@ jobs:
           - 'linux-arm64v8'
           - 'win32-ia32'
           - 'win32-x64'
+        experimental: [false]
+        include:
+          - os: macos-latest
```
lovell, line 47: Given there are no containers involved that we can use to "pin" the versions of any OS shared library dependencies, should we pin to a specific macOS release e.g. `macos-10.15`?

build/lin.sh
kleisauke: Yeah, it's a bit weird. It's intended to overwrite a Makefile variable. `RUST_TARGET_SUBDIR` is only set when cross-compiling.
https://github.com/GNOME/librsvg/blob/3661cee1cd2781cd31ac096be7e157f2e661b36b/configure.ac#L315-L319

Somehow when I enable LTO for all compiler invocations, it fails on `proc-macros`. Passing the `--target` option to Cargo seems to fix this.
https://github.com/kleisauke/libvips-packaging/blob/c22ecff4dc069f4747c6f53f4a6618a25aa59a47/build/lin.sh#L51-L55

build/lin.sh
kleisauke: Ah yes, a silly typo. Fixed with commit https://github.com/lovell/sharp-libvips/pull/44/commits/3dbab0f89b50f3126bc43928663e782847fdfae8.

build.sh
kleisauke: Good idea! Added with commit https://github.com/lovell/sharp-libvips/commit/3dbab0f89b50f3126bc43928663e782847fdfae8.

build.sh
kleisauke: It's a leftover from my early attempt to compile with Homebrew's LLVM toolchain. `llvm-strip` failed for some archive files.
https://github.com/kleisauke/libvips-packaging/runs/688284884?check_suite_focus=true#step:7:6375

In the end, adding `-fno-stack-check` didn't solve this, but switching to the Clang provided by Xcode did. I think it's a safe change to leave this in. FFmpeg seems to do the same by default for macOS 10.15, see:
http://git.videolan.org/?p=ffmpeg.git;a=commitdiff;h=07ecc4615c3ef5cc2fc61209a845cffd0c073b13

If you want, I could do a test release without this compiler flag to make sure if it's not causing any side-effects.

Regarding AVX; as fair as I know it's only conditionally enabled on runtime within libjpeg-turbo, see:
https://github.com/libjpeg-turbo/libjpeg-turbo/blob/6d8caa9f88184ad1698e91c94b37acb9506dacf0/simd/x86_64/jsimd.c#L39

.github/workflows/build-release.yml
kleisauke: Yes, lets pin it to `macos-10.15`. Commit https://github.com/lovell/sharp-libvips/pull/44/commits/3dbab0f89b50f3126bc43928663e782847fdfae8.

build/lin.sh
lovell: Got it, thanks, bit of a :exploding_head: moment there.

lovell:
Thank you, this is a great improvement that should remove many shared library conflicts/headaches. Now let's see what happens when we ask Github Actions to build us a release!

kleisauke:
Ah, I think I forgot to change something:
https://github.com/lovell/sharp-libvips/blob/master/.github/workflows/build-release.yml#L65
https://github.com/lovell/sharp-libvips/blob/master/.github/workflows/build-release.yml#L72

I'm sorry about that.

lovell:
D'oh, I didn't spot that either. I've added you as a collaborator so you can more easily fix this kind of thing ;)

kleisauke:
Commit https://github.com/lovell/sharp-libvips/commit/8f78c8177d5350dc230da2911e40d8018519a967 should fix this. Feel free to push another tag. 😅

kleisauke:
A possible improvement would be to change that check into something more constant.

```diff
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -62,14 +62,14 @@ jobs:
       run: |
         pip3 install meson ninja
         brew install advancecomp automake nasm pkg-config
-      if: matrix.os == 'macos-10.15'
+      if: matrix.platform == 'darwin-x64'
     - name: Setup toolchain (macOS)
       shell: bash
       run: |
         echo "::set-env name=CC::clang"
         echo "::set-env name=CXX::clang++"
         echo "::set-env name=MACOSX_DEPLOYMENT_TARGET::10.9"
-      if: matrix.os == 'macos-10.15'
+      if: matrix.platform == 'darwin-x64'
     - name: Build ${{ matrix.platform }}
       run: ./build.sh $(cat LIBVIPS_VERSION) ${{ matrix.platform }}
     - name: Upload Release Asset
```
(untested, maybe it doesn't work)

lovell:
Thanks Kleis, the v8.9.2-alpha2 macOS build succeeded!

It looks like the linux-armv7 build crashed - I've not seen that before, perhaps there was a temporary underlying VM problem? I'll manually copy the libvips-8.9.2-linux-armv7 file from v8.9.2-alpha1 to v8.9.2-alpha2 for now so we can get on with testing these with sharp.

kleisauke:
Hooray! The linux-armv7 build crash indeed looks strange, I think it's a temporary GitHub actions VM problem. 

I just pushed commit https://github.com/lovell/sharp/pull/2223/commits/42f0d773857f91f7d0644b8ddaa945d7fdbeaa46 to test this release.
https://travis-ci.org/github/kleisauke/sharp/builds/702318230

janakg:
hi @lovell, this is support from which release tag?

lovell:
> Prebuilt libvips binaries are now statically-linked and Brotli-compressed, requiring Node.js 10.16.0+.
> https://sharp.pixelplumbing.com/changelog#v0260---25th-august-2020

Issue #43, opened 2020-05-22
Provide Brotli compressed tarballs, ~15% smaller/faster


lovell:
I'm aware that the files provided by this repo are one of the biggest consumers of bandwidth across the whole of GitHub Releases.

Node.js 10.16.0+ natively supports Brotli decompression and initial testing suggest this would allow for 15-20% smaller release tarballs.

(Not that it matters so much, but compression time seems to be approximately the same as `advdef --shrink-insane` and decompression time is very slightly faster, although that could be simply due to less I/O.)

Things to do here:

* Ensure all the build containers have `brotli` available at the command line.
* Ensure published file extensions also include `.tar.br`

Then in sharp:

* Switch from tar to [tar-fs](https://www.npmjs.com/package/tar-fs)
* Pipe through `BrotliDecompress`, something like:

```javascript
fs
  .createReadStream(tarPath)
  .pipe(new zlib.BrotliDecompress())
  .pipe(tar.extract(vendorPath)
```

PR #46, opened 2020-06-28
Add Brotli-compressed release tarballs #43

lovell:
This will provide .tar.br files alongside the existing .tar.gz files for a given release. See #43 for more context.

@kleisauke Would net-vips benefit from these? We could also consider providing .tar.xz variants too.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -46,20 +46,31 @@ jobs:
     - name: Install dependencies (macOS)
       run: |
         pip3 install meson ninja
-        brew install advancecomp automake nasm pkg-config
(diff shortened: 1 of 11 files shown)


kleisauke:
Whoops, I just pushed https://github.com/lovell/sharp-libvips/commit/58c01b8e61063cce0b5da83d5dee2663cffd0245 to upstream instead of my fork. I wanted to make a PR instead.

This was tested with:
https://github.com/kleisauke/libvips-packaging/commit/5be3fbc499f7be1b6b78f376668d0b5d5cacd7da
https://github.com/kleisauke/libvips-packaging/actions/runs/148597283

lovell:
Thanks Kleis!

PR #48, opened 2020-07-17
Fix build failures with Rust v1.45.0

kleisauke:


lovell:
Perfect, thanks Kleis.

PR #52, opened 2020-08-09
Upgrade dep: harfbuzz

kleisauke:
And use Meson as build system. I've disabled the utils to fix the linker errors, but perhaps this should be reported upstream.

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -74,7 +74,7 @@ VERSION_GDKPIXBUF=2.40.0
 VERSION_FREETYPE=2.10.2
 VERSION_EXPAT=2.2.9
 VERSION_FONTCONFIG=2.13.92
-VERSION_HARFBUZZ=2.6.8
+VERSION_HARFBUZZ=2.7.0
 VERSION_PIXMAN=0.40.0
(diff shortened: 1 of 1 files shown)


lovell:
Thank you for working out the source of the linker errors.

kleisauke:
The patch from https://github.com/Homebrew/linuxbrew-core/pull/20910 should fix the [Linux ARMv6 build failures](https://github.com/lovell/sharp-libvips/runs/970931822), I think. Should I prepare a PR for that?

lovell:
Yes please. (I added commit 07420d5 to attempt to silence these.)

lovell:
This is being fixed upstream via https://github.com/harfbuzz/harfbuzz/pull/2628 so we could use https://patch-diff.githubusercontent.com/raw/harfbuzz/harfbuzz/pull/2628.diff as the patch.

lovell:
Let's wait for harfbuzz 2.7.1, which should be available soon. There's some discussion about the recent meson-only change at https://github.com/harfbuzz/harfbuzz/pull/2632 which suggests it was a bit rushed and did not have approval (or at least had not been reviewed).

PR #54, opened 2020-08-09
Use a minimal profile when installing Rust

kleisauke:
And install the Rust toolchain in `/usr/local/`.

This PR depends on https://github.com/lovell/sharp-libvips/pull/53.


lovell:
Thank you very much for this and all your other recent PRs Kleis.

PR #57, opened 2020-08-23
Ensure fontconfig has default config location of /etc

lovell:
I just spotted `sysconfdir` was changed in commit ee05217 but I think this value needs to be hard-coded to `/etc` for Linux. (I'm experiencing "Cannot load default config file" warnings.)

However given this script is also used for macOS, I'm unsure this will be correct e.g. homebrew uses `/usr/local/etc/`. Any thoughts @kleisauke ?

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -281,7 +281,8 @@ mkdir ${DEPS}/fontconfig
 curl -Ls https://www.freedesktop.org/software/fontconfig/release/fontconfig-${VERSION_FONTCONFIG}.tar.xz | tar xJC ${DEPS}/fontconfig --strip-components=1
 cd ${DEPS}/fontconfig
 ./configure --host=${CHOST} --prefix=${TARGET} --enable-static --disable-shared --disable-dependency-tracking \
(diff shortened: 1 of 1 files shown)


kleisauke:
Note to myself: Investigate why this doesn't work on Linux ARM64v8 (despite using the `FONTCONFIG_PATH=/etc/fonts` env variable, and installing `fonts-dejavu`). Sample run: https://app.travis-ci.com/github/kleisauke/net-vips/jobs/547767130#L250.

kleisauke:
... ignore the above comment, it seems that this is caused by an OOM error when the test suite ends (I noticed in other runs that it exits with code `137`). Perhaps I should migrate those ARM64 runners to CircleCI.

FWIW, the "Cannot load default config file" error can actually be treated as warning.
```bash
$ FONTCONFIG_FILE=doesnotexist.conf vips text x.png "Hello, World!"
Fontconfig error: Cannot load default config file
$ echo $?
0
```
(and the produced PNG file contains the famous greeting in the DejaVu font, so the default fallback config file loaded from memory works properly)

lovell:
The CircleCI "arm.medium" runners that sharp uses provide 8GB of RAM. I think TravisCI used to provide 4GB for their ARM64 environment but this might have changed recently.

PR #67, opened 2020-10-13
Add Windows arm64 support

dennisameling:
Version 8.10.2 of `build-win64-mxe` added support for Windows ARM32+ARM64: 

https://github.com/libvips/build-win64-mxe/releases/tag/v8.10.2

This PR adds support for Windows ARM64 to `sharp-libvips`. Since Microsoft is focused on ARM64 only moving forward (AFAIK), I only added that version. I'm happy to add ARM32 as well if you think that will benefit the community. 😊

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -35,7 +35,7 @@ jobs:
           - 'linux-arm64v8'
           - 'win32-ia32'
           - 'win32-x64'
-          - 'win32-arm64'
+          - 'win32-arm64v8'
         experimental: [false]
         include:
           - os: macos-10.15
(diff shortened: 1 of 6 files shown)


lovell:
Thank you very much for the PR, all looks good - please can you rebase.

We should probably consolidate the Windows-related builds into a single Dockerfile - I can do that after this PR is merged.

dennisameling:
Done, that should do the trick 😊

lovell:
Thank you, I'd prefer to keep `arm64v8` as it helps future-proof things (e.g. ARM64 v8.3 introduced JS-specific instructions) so please can you update this PR.

dennisameling:
Done! Thanks

lovell:
Brilliant, thank you Dennis.

Issue #72, opened 2020-12-18
Add linuxmusl-arm64 binaries


gpetrov:
Could you please add linuxmusl-arm64 binary generation?

Currently only linuxmusl-x64 binaries are being generated and we need musl arm64 binaries for node-alpine.

Seems to me the current build script and docker images can handle it already, just need an extra build target.

It will be a great Christmas present - now that the pubs are closed :)

lovell:
PR welcome if you're able.

Casuallynoted:
+1 to this, currently trying to build a docker image that depends on libvips but it won't compile due to the binaries not being available. Won't compile either.

gpetrov:
almost succeeded with the build for muse @lovell ! Fixed a very nasty issue with harfbuzz https://github.com/harfbuzz/harfbuzz/issues/2771

Just one final road block of some dependency mixture:
```
   Compiling glib v0.9.3
   Compiling librsvg_c_api v0.0.1 (/deps/svg/librsvg)
    Finished release [optimized] target(s) in 1m 10s
ar: `u' modifier ignored since `D' is the default (see `U')
  CCLD     librsvg-2.la
libtool:   error: object name conflicts in archive: .libs/librsvg-2.lax/librsvg_c_api.a//deps/svg/./.libs/librsvg_c_api.a
make[3]: *** [Makefile:998: librsvg-2.la] Error 1
make[3]: Leaving directory '/deps/svg'
make[2]: *** [Makefile:1277: install-recursive] Error 1
make[2]: Leaving directory '/deps/svg'
make[1]: *** [Makefile:1584: install] Error 2
make[1]: Leaving directory '/deps/svg'
```

any idea what that might be?

gpetrov:
all solved - made it optional as it seems it was included by the dependencies.

here is the PR https://github.com/lovell/sharp-libvips/pull/73

lovell:
PR #90 added this - I'm going to re-open this issue until we ship a release that uses it.

lovell:
The new linuxmusl-arm64v8 libvips binaries are now tested as part of sharp's CI process, which means prebuilt linuxmusl-arm64 sharp binaries will also be provided from the next release - see https://travis-ci.org/github/lovell/sharp/builds/761832277

Thanks all for your help with this.

PR #73, opened 2020-12-21
Linuxmusl arm64

gpetrov:
first linuxmusl arm64 build


lovell:
Thanks for the PR, I'd prefer that all the platforms/OSs provide the same set of dependencies, so we'll need to include librsvg.

Does this relate to the `object name conflicts in archive: .libs/librsvg-2.lax/librsvg_c_api.a//deps/svg/./.libs/librsvg_c_api.a` error you reported in https://github.com/lovell/sharp-libvips/issues/72#issuecomment-749093154 ? If so, that looks like a missing space, perhaps that helps you narrow down what could be causing it?

```diff
- .libs/librsvg-2.lax/librsvg_c_api.a//deps/svg/./.libs/librsvg_c_api.a
+ .libs/librsvg-2.lax/librsvg_c_api.a/ /deps/svg/./.libs/librsvg_c_api.a 
```

gpetrov:
Switched to use the standard name linuxmusl-arm64v8 (note the v8 at the end) and made alias for linuxmusl-arm64

Also adjusted your GitHub build action.

But as for the librsvg the problem is that it is just already included by other package like probably mango, so it is already available. So the dependencies are the same, but in linuxmusl somehow librsvg is already included.

The compilation seems to go all fine now. At the end is reported:
```
make[1]: Leaving directory '/deps/vips'
libvips-cpp.so.42 depends on libvips.so.42
libstdc++.so.6 does not exist in /target/lib
libc.musl-aarch64.so.1 does not exist in /target/lib
libgcc_s.so.1 does not exist in /target/lib
libstdc++.so.6 does not exist in /target/lib
libc.musl-aarch64.so.1 does not exist in /target/lib
libgcc_s.so.1 does not exist in /target/lib
     8782077     8423595  95% /packaging/libvips-8.10.0-linuxmusl-arm64v8.tar.gz
     8782077     8423595  95%
09def5d943bb4ef498138e92c3b9931016b83aaabd05fa09e3d67af6b980012d  libvips-8.10.0-linuxmusl-arm64v8.tar.br
f3b512e996e0c2f53e67f51f056a27e2cec30725ef7386c125d473711dd20cc9  libvips-8.10.0-linuxmusl-arm64v8.tar.gz
```

Wasn't sure about the missing so, files but I think it is ok as those should be available system wide.

lovell:
> as for the librsvg the problem is that it is just already included by other package like probably mango, so it is already available.

This script needs to build librsvg as a static library and cannot rely on a shared library provided via package manager. Without it I suspect libvips is being built without support for SVG input. Do you have an example build log (e.g. GitHub Actions) you could share to verify this?

gpetrov:
Here are two build logs, with and without librsvg. Second one fails. Hope you can spot the difference and have an idea for solution.

[linuxmusl-arm64v8_no_librsvg.log](https://github.com/lovell/sharp-libvips/files/5735932/linuxmusl-arm64v8_no_librsvg.log)
[linuxmusl-arm64v8_with_librsvg.log](https://github.com/lovell/sharp-libvips/files/5735936/linuxmusl-arm64v8_with_librsvg.log)

lovell:
The first log file includes the following lines that state libvips was built without support for librsvg:

```
checking for RSVG... no
configure: WARNING: librsvg-2.0 >= 2.40.3 or cairo >= 1.2 not found; disabling SVG load via rsvg
```

Please can you update this PR to include librsvg, even if there is an error, and we try to can work back from there.

gpetrov:
done it should build now with librsvg as it used to be. Maybe this will go just fine in the building actions as they are running on new instances.

I would suggest to eventually switch to docker buildx building as GitHub Actions, as it is much cleaner, like what is suggested for sqllite3:  https://github.com/mapbox/node-sqlite3/pull/1362

.github/workflows/build-release.yml
```diff
@@ -30,6 +30,7 @@ jobs:
         platform:
           - 'linux-x64'
           - 'linuxmusl-x64'
+          - 'linuxmusl-arm64v8'
```
kleisauke, line 33: Note that this target currently assumes it's running on an ARM64 device, which GitHub Actions doesn't support. AFAIK, only  [Travis supports Linux ARM64](https://docs.travis-ci.com/user/multi-cpu-architectures/) in an early beta stage, only for OSS-projects.

Perhaps we could cross-compile for `aarch64-unknown-linux-musl` on `x86_64-unknown-linux-musl` instead?

gpetrov: That is why I suggested using buildx, seems easy to do, like this sample:
https://github.com/n1ru4l/node-sqlite3/actions/runs/177214307/workflow

lovell: https://github.com/docker/setup-buildx-action appears to use QEMU for ARM - I would prefer the cross-compilation approach suggested by Kleis.

asilvas (approved):

lovell (changes requested):
Please can this be adapted to cross-compile for the `aarch64-unknown-linux-musl` host (see the existing `linux-arm64v8` directory for example settings that can be adapted).

gpetrov:
@lovell sorry for the delay, hope it is ok now.

lovell:
Thanks for the updates, I can see cross-compilation settings for rust but not for the rest of the (C-based) dependencies. Am I missing something? Perhaps we could use one of the toolchains from https://musl.cc/ ? These are available as Docker images for running on Intel hardware - `muslcc/i686:aarch64-linux-musl` might be appropriate.

build.sh
```diff
@@ -16,16 +16,25 @@ if [ $# -lt 1 ]; then
...
-  echo "- linux-arm64v8"
+  echo "- linux-arm64v8/arm64"
   echo "- darwin-x64"
   echo
   exit 1
 fi
 VERSION_VIPS="$1"
 PLATFORM="${2:-all}"
 
+# alias arm64 to arm64v8
```
lovell, line 30: Do we still need these aliases? It looks like using `linuxmusl-arm64v8` as the `PLATFORM` might be sufficient.

gpetrov:
sorry @lovell this is a bit above my head... feel free to pickup the changes further.

I'm also still getting the libtool error on librsvg.. tried about everything but can't really find a solution. It happens to me bye also on the regular linux-arm64v8

```
Compiling librsvg_c_api v0.0.1 (/deps/svg/librsvg)
    Finished release [optimized] target(s) in 1m 10s
ar: `u' modifier ignored since `D' is the default (see `U')
  CCLD     librsvg-2.la
libtool:   error: object name conflicts in archive: .libs/librsvg-2.lax/librsvg_c_api.a//deps/svg/./.libs/librsvg_c_api.a
make[3]: *** [Makefile:998: librsvg-2.la] Error 1
```

michtio:
@lovell @gpetrov is there any progress on this issue? ARM linux builds are getting more and more popular out there. I almost got all build systems running, just this one is currently a blocker with optipng-bin. 

Thanks!

lovell:
I've started a work-in-progress branch at https://github.com/lovell/sharp-libvips/tree/add-linuxmusl-arm64v8 that uses a complete cross-compiler toolchain so we can generate ARM64 binaries on x64 hardware.

Sadly it doesn't work yet because... surprise... it runs into the same original problem with librsvg that @gpetrov ran into with native compilation:

```
libtool:   error: object name conflicts in archive: .libs/librsvg-2.lax/librsvg_c_api.a//deps/svg/./.libs/librsvg_c_api.a
```

kleisauke:
This is basically what libtool does:
```bash
AR=aarch64-linux-musl-ar
f_ex_an_ar_dir=.libs/librsvg-2.lax/librsvg_c_api.a/
f_ex_an_ar_oldlib=/deps/svg/./.libs/librsvg_c_api.a

if ($AR t "$f_ex_an_ar_oldlib" | sort | sort -uc >/dev/null 2>&1); then
 :
else
  echo "object name conflicts in archive: $f_ex_an_ar_dir/$f_ex_an_ar_oldlib" >&2
fi
```

So, it checks whether the contents of the `librsvg_c_api.a` archive does not contain duplicates.

Comparing the diff between the output of `$AR t "$f_ex_an_ar_oldlib" | sort` and `$AR t "$f_ex_an_ar_oldlib" | sort -uc` reveals:
```diff
diff --git a/librsvg_c_api-sort.a b/librsvg_c_api-sort-unique.a
index 1111111..2222222 100644
--- a/librsvg_c_api-sort.a
+++ b/librsvg_c_api-sort-unique.a
@@ -164,7 +164,6 @@ floatsitf.o
 floatunditf.o
 floatunsitf.o
 fp_mode.o
-fp_mode.o
 int_util.o
 muldc3.o
 mulsc3.o
```

Notice that `fp_mode.o` is listed there twice. The reason for this is that probably one `fp_mode.o` originates from Rust and the other from musl. 

I'm looking for a way to fix this. Perhaps it needs a similar hack as the now-reverted https://github.com/GNOME/librsvg/commit/41626931b9bcc8aee5a71a735670909f70e102ac commit.

lovell:
Great work, here's the relevant logic in rustc that affects the aarch64 and musl combo:

 https://github.com/rust-lang/compiler-builtins/blob/e6fd1b272ff4cc34810e20126ffe17888a708f39/build.rs#L400-L433

kleisauke:
I noticed that too. I'm now building with `export CFLAGS_aarch64_unknown_linux_musl="+nofp"` to exclude these floating point intrinsics from Rust, hopefully that will work.

kleisauke:
This commit seems to fix it for me: https://github.com/kleisauke/libvips-packaging/commit/9f54bfdc8dec61aadc457560a98a8e6d54ea1375, see the relevant GitHub actions build:
https://github.com/kleisauke/libvips-packaging/runs/1955112764

Unfortunately, this requires a nightly version of Rust (due to the `-Zbuild-std` feature flag), perhaps the hack mentioned above would have been a better way to resolve the failing librsvg build.

lovell:
@kleisauke Thank you! I'm OK with the use of nightly rust. Feel free to add this commit to the `add-linuxmusl-arm64v8` branch, or roll it all up into a single PR, whatever is easiest.

Thanks too for improving the (rather hacked together) toolchain filesystem paths. I'll remove the offending `/usr` symlink from the `aarch64-linux-musl-crosstools-main` repo.

I'll add a native ARM64 linuxmusl-arm64v8 Travis CI environment to sharp so we can test the output. There are possible ABI differences between musl 1.1.x (Alpine 3.11, as used by all the official Node docker images) and 1.2.x (as used by the toolchain) that we might yet run into.

kleisauke:
I just opened https://github.com/lovell/sharp-libvips/pull/90. Thanks for removing the `/usr` symlink!

PR #74, opened 2020-12-24
mac: enable arm64 builds via cross compilation

abarisain:
Hi,

Xcode 12's toolchain supports cross compiling to arm64 macs. This means that we don't necessarily need a arm64 CI to build libvips for arm64 macs.

This PR adds support for a new target, `darwin-arm64v8`, which builds for Apple Silicon. Running `build.sh 8.10.15 darwin-arm64v8` produces arm64 dylibs.

I have not tested the library extensively into a project.
I also haven't tried building this target on a M1 mac (even though it would probably work) as this is not my goal just yet.

This PR is essentially a proof of concept to see if it is something you'd like pursuing before I invest more work into it.

Compromises:
- Some "if $DARWIN_ARM" here and there
- ~~Skip libffi, meaning that it requires minor patching of meson projects due to being stuck on 0.55. The libffi built with the python script is broken on arm64, fortunately the meson version works great~~ UPDATE: Patched it
- ~~Removed rsvg for the POC (I'm aware that you'd like to keep the libvips build the same on all platforms, so this is only temporary). Rust is a blocker on this one, unless we downgrade rsvg (but that needs building another dependency)~~ UPDATE: Added rustup which downloads a beta rust

Otherwise, I find it pretty clean. Kudos to you for making platforms so easy to add in the build script!

Happy holidays,


abarisain:
Oops, the PR also adds `VERSION_HARFBUZZ=2.7.3` which was required for me to build. I'll keep it there for the time being so that you can easily pull the branch and build it.

This should be updated before merging this

abarisain:
Fixed a part of the script that I forgot to commit (PKG_CONFIG_PATH) which broke the build

Added rustup, used it to download a beta rust for arm64. I think that macx64 builds should also be using rustup is better as the version of rust (and the installed toolchains) can be controlled without relying on what's in the image

lovell:
Merci beaucoup for making a start on this, it is definitely something worth pursuing.

It probably makes sense to wait for meson v0.56.1 (which also contains fixes for a couple of regressions in v0.56.0 relating to pkg-config and static libraries) and rust v1.49 as these should make things a little simpler.

I've just updated harfbuzz in the master branch so you should be able to rebase and pick up this change.

.github/workflows/build-release.yml
```diff
@@ -48,7 +50,7 @@ jobs:
       run: |
         pip3 install meson==0.55.3 ninja
         brew install advancecomp automake brotli nasm pkg-config
-      if: matrix.platform == 'darwin-x64'
+      if: matrix.platform == 'darwin-x64' || matrix.platform == 'darwin-arm64v8'
```
lovell, line 53: Might be able to use something like `if: contains(matrix.platform, 'darwin')` here.

abarisain: Ah, thanks. I'm quite unfamiliar with the github workflow syntax!

abarisain:
>It probably makes sense to wait for meson v0.56.1 (which also contains fixes for a couple of regressions in v0.56.0 relating to pkg-config and static libraries) and rust v1.49 as these should make things a little simpler.

Rust 1.49 is definitely something we should wait for, as it only comes out in 6 days.

I do believe that taking control of the rust version with rustup makes a for a bit more complicated, but more robust script. That said, it's 100% your decision and I don't feel super strongly about it.

Meson, yeah that will clean up the build script nicely. The meson tweaks are way too hackish for my taste.

Anyhow, I've used the built binary with success (though we don't use all image formats, so ymmw). While this PR is still a WIP, anyone who needs a M1 compatible sharp can pull this branch, build it and serve the files over http. As sharp seems to cache in in .npm, it's fine enough for my uses! No rush to merge on my side :)

Feel free to comment on any of my changes when you've got time, I'll be happy to tweak the PR.

lovell:
Switching the macOS build to use rustup would make it consistent with the Linux environments so I'm happy with this improvement.

abarisain:
Updated to use stable rust

And what do you know, meson 0.56.1 came out today! I'll update the PR and see if I can remove some of the hacks

abarisain:
Removed meson hacks

Both x64 and arm64 build fine. Haven't tested that the built libraries work yet

As far as I am concerned, this PR isn't WIP anymore !

PS: had to update halfbuzz once again

build/lin.sh
```diff
@@ -280,6 +303,10 @@ sed -i'.bak' "/\[ 'xpm'/{N;N;N;N;d;}" gdk-pixbuf/meson.build
 # Ensure meson can find libjpeg when cross-compiling
 sed -i'.bak' "s/has_header('jpeglib.h')/has_header('jpeglib.h', args: '-I\/target\/include')/g" meson.build
 sed -i'.bak' "s/cc.find_library('jpeg'/dependency('libjpeg'/g" meson.build
+if [ $DARWIN_ARM = true ]; then
```
lovell, line 306: Can we remove the conditional and make this change for everything?

build/lin.sh
```diff
@@ -155,12 +175,15 @@ cd ${DEPS}/zlib
 ./configure --prefix=${TARGET} ${LINUX:+--uname=linux} ${DARWIN:+--uname=darwin} --static
 make install
 
+# Do not build ffi on darwin-arm64v8 so that meson will build it itself properly
+if [ $DARWIN_ARM != true ]; then
```
lovell, line 179: This appears to make meson fall back to a "wrapped" https://gitlab.freedesktop.org/gstreamer/meson-ports/libffi fork, which is not quite the same as libffi v3.3 and also prevents control over the `--disable-*` build flags we're specifying. Is this condition still required?

lovell (commented):
Thank you very much for the updates, plus I noticed you'd rebased against all the latest changes too. I've left a couple of questions inline but otherwise I think this is almost ready to merge.

build/lin.sh
abarisain: We could. I didn't want to impact other platforms with a change that's only really needed because I'm skipping libffi.

I'll hold off making this change until I give fixing libffi a shot.

build/lin.sh
abarisain: > also prevents control over the --disable-* build flags 

Oh, I totally forgot about those "--disable". That's a problem.

Making meson fall back on a wrapped libffi fork is by design, as the mesonified libffi happens to build properly while trunk libffi does not.
Unfortunately, libffi didn't tag a release in a long while, so the condition is still required.

That said, now that I know you're interested in the PR, I'll look into solutions! I'm pretty sure the homebrew/macports people have working buildscripts.

abarisain:
I copied homebrew's libffi patch, which fixes the build. I removed the meson sed as a result

Now I just need to rebase this on top of master

build/lin.sh
abarisain: Removed the patch

build/lin.sh
```diff
@@ -140,6 +153,13 @@ if [ "$ALL_AT_VERSION_LATEST" = "false" ]; then exit 1; fi
 
 # Download and build dependencies from source
 
+if [ $DARWIN = true ]; then
+  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --no-modify-path -y
+  if [ $DARWIN_ARM = true ]; then
```
kleisauke, line 158: Variable needs to be quoted before comparison.
```suggestion
  if [ "$DARWIN_ARM" = true ]; then
```

build/lin.sh
```diff
@@ -158,6 +178,11 @@ make install
 mkdir ${DEPS}/ffi
 $CURL https://github.com/libffi/libffi/releases/download/v${VERSION_FFI}/libffi-${VERSION_FFI}.tar.gz | tar xzC ${DEPS}/ffi --strip-components=1
 cd ${DEPS}/ffi
+if [ $DARWIN_ARM = true ]; then
```
kleisauke, line 181: Variable needs to be quoted before comparison.
```suggestion
if [ "$DARWIN_ARM" = true ]; then
```

kleisauke (commented):
I've left a couple of inline suggestions, but otherwise this looks good to me.

abarisain:
Thanks, I'll apply them! I must admit that I'm not good at bash.

There are merge conflicts, do you want me to fix them?

kleisauke:
Feel free to fix them, sorry for causing merge conflicts. I had a look at how this behaves in CI, see:
https://github.com/abarisain/sharp-libvips/compare/darwin-arm64...kleisauke:darwin-arm64

It took 2 commits to fix the failing CI builds, see:
| Failing build | Fix |
| ---------------- | ---- |
| https://github.com/kleisauke/libvips-packaging/runs/1777336006 | https://github.com/kleisauke/libvips-packaging/commit/f3efa320eb4661ce8b4f710e9c79c8e2622a4554 |
| https://github.com/kleisauke/libvips-packaging/runs/1777505940 | https://github.com/kleisauke/libvips-packaging/commit/42c0de446e553e02c3dac2601f1949fe766831c6  |

(Also, the `-target arm64-apple-darwin` flag seems no longer necessary, so I removed it with commit https://github.com/kleisauke/libvips-packaging/commit/5fc7ff3f4c6bb6df68a042d744a02a54f4d8102e)

I remembered what a hassle it was to support `darwin-x64` in the first place, but this seems to be very straightforward. Thanks!

Anyways, here are the produced libraries:
https://github.com/kleisauke/libvips-packaging/releases/tag/v8.10.5-darwin-arm64

This can be tested on a ARM-based macOS with:
```bash
SHARP_DIST_BASE_URL="https://github.com/kleisauke/libvips-packaging/releases/download/v8.10.5-darwin-arm64/" \
  npm install sharp
```

abarisain:
>Feel free to fix them, sorry for causing merge conflicts

It's quite alright, this happens all the time!

Thanks for fixing the CI and providing a dist URL.

I don't think there is anything left for me to fix, your commits look good: I pulled them in this PR. Sorry you had to cleanup after me.

>Also, the -target arm64-apple-darwin flag seems no longer necessary

Nice catch, the meson fixes removed the need for this.

>I remembered what a hassle it was to support darwin-x64 in the first place, but this seems to be very straightforward. Thanks!

I thank your script's modularity and architecture that made this easy with little changes. Apple also really made cross compilation quite straightforward.
I'm lucky all parts (meson, rust) fell into place at just the right time.

abarisain:
Ah my bad, had to merge a file. I did not have the chance to test this build as I can't access my intel mac right now. CI should take care of this, right?

I did a merge commit rather than a rebase, if that's fine by you

lovell:
Please can you rebase against the `upstream/master` branch to pick up a CI change that verifies the build is good as part of the PR process.

abarisain:
Done!

I haven't had the chance to build this locally yet, but the CI should do it.

abarisain:
A nice enhancement would be to have a post build script that ensures that the darwin builds are of the expected architecture using lipo/file.

lovell:
Merci beaucoup, CI improvements are always welcome. Ultimately this repo should try to run the latest sharp unit tests against the resultant binaries, although there will be some changes that require updates to both repos.

abarisain:
Thanks for reviewing and getting this merged!
If something breaks/changes and you need someone with a M1 mac to test/help, feel free to ping me. 

I'll see about adding the architecture checks in another PR.

Unfortunately running sharp tests will require an arm64 mac CI and there's no saying when that will be available.

PR #76, opened 2021-01-13
Use stable version of glib

inukshuk:
I noticed that the build is using the glib development version. In case this is not the intention, here is a PR that switches to 2.66.4, [the current stable release](https://developer.gnome.org/glib/). Unfortunately release-monitoring reports the latest dev, so I disabled the version check.

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -69,7 +69,7 @@ CURL="curl --silent --location --retry 3 --retry-max-time 30"
 # Dependency version numbers
 VERSION_ZLIB=1.2.11
 VERSION_FFI=3.3
-VERSION_GLIB=2.67.2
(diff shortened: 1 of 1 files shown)


lovell:
Is there a specific problem with glib v2.67.2 you've run into? (The use of "unstable" in the context of glib generally refers to patch version changes to new, unused API rather than stability of existing APIs.)

inukshuk:
No, I did not run into any issues specific to v2.67.2. I assumed that you were probably using this version intentionally, but I just noticed the designation as 'development version' and the fact that 2.66.4 is the latest version on Arch Linux.

(I'm having unrelated issues with the statically linked build on Linux with Electron, but I have not tracked down the root cause yet.)

PR #80, opened 2021-01-26
glib: use patch from Alpine Linux

kleisauke:
Alpine Linux's patch seems to be compatible again with the latest GLib versions. I also switched to using the new Meson configure option for disabling the tests.


lovell:
Lovely stuff.

PR #81, opened 2021-01-27
Build a single shared library

kleisauke:
This PR links `libvips.so.42` statically into `libvips-cpp.so.42` as described in issue https://github.com/lovell/sharp/issues/2535. It also ensures that it can interoperate with a system-wide installed GLib by un-exporting the `g_param_spec_types` symbol on Linux.

On macOS the build script remains basically the same except that `libvips.42.dylib` is also statically linked into `libvips-cpp.42.dylib`.

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -395,14 +395,20 @@ make install-strip
 mkdir ${DEPS}/vips
 $CURL https://github.com/libvips/libvips/releases/download/v${VERSION_VIPS}/vips-${VERSION_VIPS}.tar.gz | tar xzC ${DEPS}/vips --strip-components=1
 cd ${DEPS}/vips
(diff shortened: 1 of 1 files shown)


lovell:
Looks great, thank you Kleis. I'm going to publish a patch release of sharp first, then I'll come back to this and the related sharp PR later, probably at the weekend.

There are a couple of improvements relating to JPEG and PNG error handling in the forthcoming libvips v8.10.6 so I think we should probably wait for that as part of this improvement.

lovell:
Please can you rebase to pick up https://github.com/lovell/sharp-libvips/commit/75e522930531f5af48810032988dd3bf721d1e90 to ensure the build is still good then we can get this merged, thank you.

lovell:
Brilliant, thank you Kleis.

PR #82, opened 2021-01-27
Revert "glib: use patch from Alpine Linux (#80)"

kleisauke:
The Alpine patch seems to apply without problems, but further on it causes issues when configuring GLib, see for example:
https://github.com/kleisauke/libvips-packaging/runs/1777505762

Sorry for not testing this properly in the first place.


lovell:
Thanks, no worries.

PR #86, opened 2021-02-02
Build linux-arm64v8 with Ubuntu Bionic

milkcocoa:
As you know, the Arm64 based Node.js official Docker image is built with Debian 10 buster, or 9 stretch whose glibc versions are  less than 2.29, so it is hard to use the pre-built sharp npm package in these environments.

According to https://github.com/lovell/sharp/issues/1651 and https://github.com/lovell/sharp-libvips/commit/2e4202d671904666c012b142b8e21be017d30a28 discussion, gcc 8.2.0 has a bug, but 9.x doesn't, and gcc 8.4.0 seems to have fixed it.

Ubuntu 18.04 Bionic provides the gcc 8.4 through `g++-8-aarch64-linux-gnu` package, and I’ve tried to build it with my M1 MacBook Pro. Though it failed to build the `libsvg` with the rust error, I suppose it can be done with the Intel machine, because the error is the same one discussed in https://github.com/rust-lang/rust/issues/80346, so at this time I’d like to discuss the possibility of this approach.

Unfortunately, since Ubuntu 16.04 Xenial doesn’t have the `g++-8-aarch64-linux-gnu` package, the minimum glibc version would be 2.27, and Debian 9 stretch cannot be supported.


linux-arm64v8/Dockerfile
```diff
@@ -9,10 +9,20 @@ ENV \
...
   apt-get update && \
-  apt-get install -y curl && \
+  apt-get install -y \
+    curl \
+    apt-transport-https \
+    gnupg \
+    software-properties-common \
+    && \
+  curl https://apt.kitware.com/keys/kitware-archive-latest.asc | apt-key add - && \
+  apt-add-repository 'deb https://apt.kitware.com/ubuntu/ bionic main' && \
```
milkcocoa, line 24: The latest version CMake is necessary to build `aom` with `make install/strip`

lovell:
Konnichiwa, thank you for the PR.

I note this is still a WIP - have you been able to successfully use gcc 8.4.0 running on x64 to cross-compile a libvips binary that passes all the sharp unit tests when run natively on ARM64?

(I'm currently in the process of securing dedicated ARM64 hardware to help verify this kind of thing as part of the CI process.)

milkcocoa:
こんにちは (Konnichiwa) Thanks for replying.

Unfortunately, I did not succeed in compiling `libvips` because of Rust error in `libsvg` caused by the QEMU bug in Docker with my M1 MacBook Pro.
I wanted to ask you whether it is OK for you to use older Ubuntu, and gcc-8.4.

After creating this pull request, instead of using amd64 based Docker image, I've tried to use the arm64v8 based image, and succeeded in compiling both `libsvg`, and `libvips`.
I've attached the build artifact here.

You can see the branch from the following URL.
https://github.com/milkcocoa/sharp-libvips/tree/linux-arm64v8-bionic-arm-based

I'm currently not sure how to perform the sharp unit tests for the build artifact in my macOS environment.

[libvips-8.10.5-linux-arm64v8.tar.gz](https://github.com/lovell/sharp-libvips/files/5938926/libvips-8.10.5-linux-arm64v8.tar.gz)

lovell:
My original comment about 8.4.0 was only a guess, so I've done some digging through the gcc changelog and found the commit that I believe fixes the problem - see https://github.com/gcc-mirror/gcc/commit/44a7251e1e97caf3b2315f20d48519ff8fced152 - this was first made available in gcc 9.1.0 and was not backported any further, so sadly I don't think the use of gcc 8 will help us.

lovell:
I'll close this for now as sadly I don't think this approach will get us the gcc 9 we need. Thank you very much for the PR, and if you or anyone else finds a way to easily solve this, please do feel free to open another PR or issue to discuss.

PR #88, opened 2021-02-19
Reduce size of binaries by up to 10%

lovell:
This PR allows the size vs perf optimisation level to be set per dependency.

There's no change (`-O3`) for vips, gif, orc, webp, spng, png16, jpeg, heif, lcms2, zlib, aom; everything else is optimised for size (`-Os`). This accounts for 7-9% reduction in binary size.

For Linux, it adds flags that allows the linker to remove unused sections (unsupported by cairo). This accounts for 1% reduction in binary size.

For Linux x64, it sets a minimum CPU requirement of Westmere (circa 2010) or later. This allows gcc to use CLMUL intrinsics, which helps improve zlib (de)compression performance (whilst we wait for zlib-ng).

@kleisauke Is there anything you think could be added/improved? Perhaps this approach might be something to consider for the Windows builds too?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -47,7 +47,7 @@ for flavour in darwin-x64 darwin-arm64v8; do
     # See:
     # https://forums.developer.apple.com/thread/121887
     # https://trac.ffmpeg.org/ticket/8073#comment:12
-    export FLAGS="-O3 -fPIC -fno-stack-check"
+    export FLAGS="-fno-stack-check"
 
     if [ $PLATFORM = "darwin-arm64v8" ]; then
       # ARM64 builds work via cross compilation from an x86_64 machine
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -24,6 +24,15 @@ esac
 mkdir ${DEPS}
 mkdir ${TARGET}
 
+# Default optimisation level is for binary size (-Os)
+# Overriden to performance (-O3) for select dependencies that benefit
+export FLAGS+=" -Os -fPIC"
+
+# Allow linker to remove unused sections
+if [ "$LINUX" = true ]; then
+  export FLAGS+=" -ffunction-sections -fdata-sections"
+fi
+
 # Common build paths and flags
 export PKG_CONFIG_LIBDIR="${TARGET}/lib/pkgconfig"
 export PATH="${PATH}:${TARGET}/bin"
@@ -39,7 +48,7 @@ export LDFLAGS="-L${TARGET}/lib"
 # On Linux, we need to create a relocatable library
 # Note: this is handled for macOS using the `install_name_tool` (see below)
 if [ "$LINUX" = true ]; then
-  export LDFLAGS+=" -Wl,-rpath='\$\$ORIGIN/'"
+  export LDFLAGS+=" -Wl,--gc-sections -Wl,-rpath='\$\$ORIGIN/'"
 fi
 
 # On macOS, we need to explicitly link against the system libraries
@@ -178,7 +187,7 @@ fi
 mkdir ${DEPS}/zlib
 $CURL https://zlib.net/zlib-${VERSION_ZLIB}.tar.xz | tar xJC ${DEPS}/zlib --strip-components=1
 cd ${DEPS}/zlib
-./configure --prefix=${TARGET} ${LINUX:+--uname=linux} ${DARWIN:+--uname=darwin} --static
+CFLAGS="${CFLAGS} -O3" ./configure --prefix=${TARGET} ${LINUX:+--uname=linux} ${DARWIN:+--uname=darwin} --static
 make install
 
 mkdir ${DEPS}/ffi
@@ -232,7 +241,7 @@ make install-strip
 mkdir ${DEPS}/lcms2
 $CURL https://downloads.sourceforge.net/project/lcms/lcms/${VERSION_LCMS2}/lcms2-${VERSION_LCMS2}.tar.gz | tar xzC ${DEPS}/lcms2 --strip-components=1
 cd ${DEPS}/lcms2
-./configure --host=${CHOST} --prefix=${TARGET} --enable-static --disable-shared --disable-dependency-tracking
+CFLAGS="${CFLAGS} -O3" ./configure --host=${CHOST} --prefix=${TARGET} --enable-static --disable-shared --disable-dependency-tracking
 make install-strip
 
 mkdir ${DEPS}/aom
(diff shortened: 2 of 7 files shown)


kleisauke (approved):
I noticed that aom is built with `-Os` and the PR description mentions that it's built with `-O3`, is this intended? Otherwise, this looks good to me.

Regarding the Windows builds, I tried this in the past but removed these flags (`-fdata-sections -ffunction-sections` and `-Wl,--gc-sections`) with commit [libvips/build-win64-mxe@1964b87](https://github.com/libvips/build-win64-mxe/commit/1964b876290d151c124abd80902a7b12fe464184#diff-d6398ba75cd0ddee320bb4ca6fbd7e1aa2ea8d2b952589e14af0f68ff684cc7d) since the binaries ended up being smaller _without_ these flags. I should probably try again with setting `-Os` per dependency that might benefit from it.

fwiw, aom is the largest dependency (with a size of 7.9 MB) included in the Windows builds, so that could be a good candidate. Alternatively, switching to the dav1d+rav1e combo might also help (see the [`dav1d-rav1e`](https://github.com/libvips/build-win64-mxe/tree/dav1d-rav1e) branch), since these are combined only 4.2 MB in size.

lovell:
Thanks Kleis, great feedback.

It's interesting you saw a size increase with the use of `--gc-sections` when cross-compiling. I saw a decrease with the native builds but didn't look so closely at the ARM targets so will double-check those now.

I meant to add in the original description that the aom [CMakeLists.txt](https://aomedia.googlesource.com/aom/+/refs/heads/master/CMakeLists.txt#368) suggests it follows the [CMAKE_BUILD_TYPE](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) convention where the default is `-O3` unless you specify `MinSizeRel` to imply `-Os`, but we should check this too.

Happy to switch to a dav1d+rav1e combo if we can get it working on all platforms - let's do that in a separate PR.

lovell:
I verified the cross-compiled ARM binaries and they are smaller with `--gc-sections`.

kleisauke:
> It's interesting you saw a size increase with the use of `--gc-sections` when cross-compiling.

Support for `--gc-sections` on PE/COFF binaries was landed in binutils 2.26, and at the time I was testing with 2.28 (which is already quite old). I'll do some testing with the latest binutils version, perhaps that could help.

> I meant to add in the original description that the aom [CMakeLists.txt](https://aomedia.googlesource.com/aom/+/refs/heads/master/CMakeLists.txt#368) suggests it follows the [CMAKE_BUILD_TYPE](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) convention where the default is `-O3` unless you specify `MinSizeRel` to imply `-Os`, but we should check this too.

It seems that the `-O3` compiler flag in aom is forced for Emscripten only. I think CMake always follows the `CFLAGS` environment variable, unless the `CMAKE_C_FLAGS` variable is explicitly defined.

lovell:
Thanks, let's go for dav1d and rav1e to replace aom.

I spotted a new `strip` feature in cargo nightly/unstable that will soon help reduce the size of rust-based dependencies.

https://doc.rust-lang.org/cargo/reference/unstable.html#profile-strip-option

```sh
export CARGO_PROFILE_RELEASE_STREAM=symbols
```

kleisauke:
Here's a size comparison for the Windows binaries using different toolchains and different flags.

**Toolchains used**
- GCC 10.2.0 / binutils 2.36.1
- LLVM 11.1.0

**Commits tested**
| CFLAGS | LDFLAGS | Commit |
| ------ | ------- | ------ |
| -O3 | -Wl,-s | https://github.com/libvips/build-win64-mxe/commit/e3e5bfd0e57e79451a28dae67271bee2891b3305 |
| -O3 -fdata-sections -ffunction-sections | -Wl,--gc-sections -Wl,-s | https://github.com/libvips/build-win64-mxe/commit/98f08b739d191215ce61632d5f27ce23949e9e7c |
| -O3 -flto | -flto -Wl,-s | https://github.com/libvips/build-win64-mxe/commit/9fffbad48f418aa3b92b4138c3cd57b563fa2b2d |

**Comparison table**
<table>
<tr>
<th>Toolchain</th>
<th>CFLAGS</th>
<th>LDFLAGS</th>
<th>Size of binaries</th>
</tr>
<tr>
<td>GCC</td>
<td>-O3</td>
<td>-Wl,-s</td>
<td>
<pre lang="bash">
1.6M    libglib-2.0-0.dll
356K    libgobject-2.0-0.dll
23M     libvips-42.dll
992K    libvips-cpp-42.dll
26M     total
</pre>
</td>
</tr>
<tr>
<td>GCC</td>
<td>-O3 -fdata-sections -ffunction-sections</td>
<td>-Wl,--gc-sections -Wl,-s</td>
<td>
<pre lang="bash">
1.6M    libglib-2.0-0.dll
356K    libgobject-2.0-0.dll
25M     libvips-42.dll
972K    libvips-cpp-42.dll
28M     total
</pre>
</td>
</tr>
<tr>
<td>GCC</td>
<td>-O3 -flto</td>
<td>-flto -Wl,-s</td>
<td>
<pre lang="bash">
1.7M    libglib-2.0-0.dll
352K    libgobject-2.0-0.dll
24M     libvips-42.dll
984K    libvips-cpp-42.dll
27M     total
</pre>
</td>
</tr>
<tr>
<td>LLVM</td>
<td>-O3</td>
<td>-Wl,-s</td>
<td>
<pre lang="bash">
1.5M    libglib-2.0-0.dll
284K    libgobject-2.0-0.dll
21M     libvips-42.dll
928K    libvips-cpp-42.dll
23M     total
</pre>
</td>
</tr>
<tr>
<td>LLVM</td>
<td>-O3 -fdata-sections -ffunction-sections</td>
<td>-Wl,--gc-sections -Wl,-s</td>
<td>
<pre lang="bash">
1.5M    libglib-2.0-0.dll
284K    libgobject-2.0-0.dll
19M     libvips-42.dll
484K    libvips-cpp-42.dll
21M     total
</pre>
</td>
</tr>
<tr>
<td>LLVM</td>
<td>-O3 -flto</td>
<td>-flto -Wl,-s</td>
<td>
<pre lang="bash">
1.7M    libglib-2.0-0.dll
340K    libgobject-2.0-0.dll
20M     libvips-42.dll
932K    libvips-cpp-42.dll
23M     total
</pre>
</td>
</tr>
</table>

So it seems that with GCC, the size optimization flags have no effect because it makes the binaries larger. On the other hand, LLVM works well here, so we might consider switching to this toolchain for building the Windows binaries.

/cc @mstorsjo as he might be interested in this comparison.

lovell:
@kleisauke Great research, thank you, I agree with switching in this case. I wonder what LLVM with gc was able to do that almost halved the size of `libvips-cpp-42.dll`? Perhaps it's better at breaking source files up into smaller compilation units?

PR #89, opened 2021-02-28
Switch from libjpeg-turbo to mozjpeg

lovell:
libvips resets to libjpeg-turbo's `JCP_FASTEST` defaults so this can be considered a "no-op" unless mozjpeg features are explicitly set.

https://github.com/libvips/libvips/blob/dd57463c9a7f80730d70c4ff080bbd3f5ab1cc80/libvips/foreign/vips2jpeg.c#L594-L600

@kleisauke Would you like me to submit a PR to make this change as the default to https://github.com/libvips/build-win64-mxe/ also?


kleisauke (approved):
Looks good to me! https://github.com/libvips/build-win64-mxe/pull/21 will do this for the Windows builds.

PR #90, opened 2021-03-01
Add linuxmusl-arm64v8 cross-compiler environment

kleisauke:
Resolves https://github.com/lovell/sharp-libvips/issues/72, and closes https://github.com/lovell/sharp-libvips/pull/73.

(The CI job was marked as experimental since this environment uses a nightly version of Rust)


lovell:
Wonderbaar!

kleisauke:
It looks like this target depends on a shared `libintl.so`, see:
```
Library intl found: NO

|Executing subproject proxy-libintl method meson 
|
|Project name: proxy-libintl
|Project version: 1
|C compiler for the host machine: /bin/aarch64-linux-musl-gcc (gcc 10.2.1 "aarch64-linux-musl-gcc (GCC) 10.2.1 20210116")
|C linker for the host machine: /bin/aarch64-linux-musl-gcc ld.bfd 2.35
|C compiler for the build machine: cc (gcc 9.3.0 "cc (Alpine 9.3.0) 9.3.0")
|C linker for the build machine: cc ld.bfd 2.33.1
|Build targets in project: 2
|Subproject proxy-libintl finished.
<snip />
Installing subprojects/proxy-libintl/libintl.so to /target/lib
Stripping target 'subprojects/proxy-libintl/libintl.so' using /bin/aarch64-linux-musl-strip.
<snip />
libvips-cpp.so.42 depends on libintl.so
<snip />
-rwxr-xr-x    1 root     root          9920 Mar 26 21:05 libintl.so
```
(Noticed this while debugging a [Travis failure](https://travis-ci.org/github/kleisauke/net-vips/jobs/765091993#L150), sorry for not having caught this earlier).

I think the easiest way to fix this is to pass `-Dproxy-libintl:default_library=static` while configuring GLib.
https://mesonbuild.com/Release-notes-for-0-54-0.html#per-subproject-default_library-and-werror-options

But this made me thinking, perhaps we could just use [`proxy-libintl`](https://github.com/frida/proxy-libintl) instead of `gettext-runtime` for all these targets (similar to the Windows builds, see https://github.com/libvips/libvips/issues/1637#issuecomment-629672415)? I could open a new issue if you want.

lovell:
No worries, I guess/hope most people using Alpine Linux will already have installed `libintl` given there are [435 packages dependent upon it](https://pkgs.alpinelinux.org/package/v3.13/main/x86_64/libintl). The use of `proxy-libintl` definitely makes sense here and it should be suitable (with `-DSTUB_ONLY`) for all the musl and macOS targets.

kleisauke:
I had an attempt to fix the CI failures on the [`use-proxy-libintl`](https://github.com/lovell/sharp-libvips/tree/use-proxy-libintl) branch:
https://github.com/lovell/sharp-libvips/compare/use-proxy-libintl...kleisauke:use-proxy-libintl

It seems to work (the last 2 commits should probably be landed separately).

lovell:
Brilliant, thank you for working that out, if it's OK I'll cherry pick the relevant commits and open a PR.

lovell:
I've cherry-picked (and squashed a couple of) your commits onto the `master` branch so hopefully https://github.com/lovell/sharp-libvips/commits/master looks about right.

kleisauke:
Thanks!

PR #91, opened 2021-03-01
Add libimagequant v2.4.1 BSD-2-Clause

lovell:
Although I've done a fair amount of testing with it locally, this should probably be considered _experimental_ and I'll be marking it as such in the sharp documentation.

Due to the use of C99 variable length arrays, MSVC is totally unsupported. If gcc >= 9 is found then OpenMP support is enabled (clang will require https://github.com/mesonbuild/meson/issues/7435).

More details at https://github.com/lovell/libimagequant

### DIFF
--- a/THIRD-PARTY-NOTICES.md
+++ b/THIRD-PARTY-NOTICES.md
@@ -21,6 +21,7 @@ used under the terms of the following licences:
 | libffi        | MIT Licence                                                                                               |
 | libgsf        | LGPLv3                                                                                                    |
 | libheif       | LGPLv3                                                                                                    |
+| libimagequant | [BSD-3-Clause](https://github.com/lovell/libimagequant/blob/main/COPYRIGHT)                               |
 | mozjpeg       | [zlib License, IJG License, BSD-3-Clause](https://github.com/mozilla/mozjpeg/blob/master/LICENSE.md)      |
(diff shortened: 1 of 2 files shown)


kleisauke (approved):
LGTM! Just wondering, does libimagequant perform much better with OpenMP enabled? The Windows builds have [disabled OpenMP in GCC](https://github.com/libvips/build-win64-mxe/blob/d7b8c8dccb48745861cdf40ab35d4a3d6f63691b/build/plugins/gcc/overrides.mk#L13-L19), as only *magick was using that, and I wasn't sure if there was a big advantage to have that enabled.

lovell:
Use of OpenMP is only used for larger images, e.g. there's a >= 3Mpx conditional on the remap loop.

https://github.com/lovell/libimagequant/blob/55e896fe6e0d5bb379acefe6b974b20f364706e3/libimagequant.c#L1002

kleisauke:
I just noticed that the Linux binaries will depend upon a system-wide libgomp after this PR, see:
```bash
libgomp.so.1 does not exist in /target/lib
```
I'm not sure if we can assume that this library is always present.

lovell:
Good spot, no we can't. Given this is experimental I'll make OpenMP optional in https://github.com/lovell/libimagequant and default to false.

lovell:
Done. I've force-updated the v2.4.1 tag so the next build will pick it up automagically.

PR #92, opened 2021-03-03
Build a more minimal libxml2

lovell:
Rather than exclude some modules that are not required, this approach excludes all optional modules then includes only what is needed.

This reduces the binary size by ~0.3MB.

When https://gitlab.gnome.org/GNOME/libxml2/-/commit/106757e8c1e26ad9b8c924c7f304074b79e082c5 is released we can also remove the `--with-regexps` flag.

@kleisauke We might want to use this approach for the Windows builds too.

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -216,9 +216,10 @@ ninja -C _build install
 mkdir ${DEPS}/xml2
 $CURL http://xmlsoft.org/sources/libxml2-${VERSION_XML2}.tar.gz | tar xzC ${DEPS}/xml2 --strip-components=1
 cd ${DEPS}/xml2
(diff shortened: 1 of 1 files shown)


kleisauke (approved):
LGTM! It's a bit unfortunate that this dependency is only needed for libgsf and librsvg ([libgsf can possibly be replaced by miniz](https://github.com/libvips/libvips/pull/465#issuecomment-447886980) and [I've just asked librsvg if libexpat is ever considered](https://gitlab.gnome.org/GNOME/librsvg/-/issues/224#note_1050670)).

I'll incorporate this change into the Windows builds for the next libvips release.

PR #94, opened 2021-03-06
aom: build without HDR/WebM support, 13% smaller macOS binaries

lovell:
Linux x64 before/after:
```
23738256 libvips-cpp.so.42.12.6
22792080 libvips-cpp.so.42.12.6
```

macOS x64 before/after:
```
26306752 libvips-cpp.42.dylib
22801512 libvips-cpp.42.dylib
```


kleisauke (approved):

lovell:
@kleisauke Shall I open a PR with this change the Windows builds too?

kleisauke:
Sure, that'd be great!

PR #96, opened 2021-03-06
Switch from zlib to zlib-ng (stable release candidate)

### LINKED ISSUE #25: Investigate SIMD-accelerated zlib implementations
It might be worth switching to a performance-oriented fork of the popular zlib library within the prebuild libvips binaries. Several companies have made zlib significantly faster by using modern CPU instructions (SSE, ARM NEON).

The [zlib-ng project](https://github.com/zlib-ng/zlib-ng) seems the most promising, it has merged the changes of both [Cloudflare](https://github.com/cloudflare/zlib) and [Intel](https://github.com/jtkukunas/zlib ) forks and cleans up the code in general. It's API / ABI compatible with zlib (when configured with `--zlib-compat` or `-DZLIB_COMPAT=ON`), so it can be used as a drop-in replacement (just like libjpeg-turbo).

There is no "stable" release for zlib-ng yet, so it might be too risky to migrate now. However, it's already been continuously fuzzed by OSS-Fuzz<sup>[1]</sup> and the Chromium / AOSP developers are also interested in switching to this fork<sup>[2][3]</sup>.

I've already done some experiments with zlib-ng within the MXE-based Windows builds<sup>[4]</sup>, and see a nice performance improvement by converting a 4,000 x 5,000 JPEG image to PNG with different compression values<sup>[5]</sup>.

<sup>[1]</sup> https://github.com/google/oss-fuzz/tree/master/projects/zlib-ng
<sup>[2]</sup> https://crbug.com/687631
<sup>[3]</sup> https://groups.google.com/a/chromium.org/d/msg/blink-dev/T3rPfd6G6ps/l8pAk7gnBgAJ
<sup>[4]</sup> https://github.com/libvips/build-win64-mxe/tree/zlib-ng-experiment
<sup>[5]</sup> https://gist.github.com/kleisauke/879e3075a675c6f945dff3772d2ef0a3

lovell:
The only changes to zlib-ng since RC2 appear to relate to tests, docs and CI so I'm happy it will be very close to the (supported) release, which means it's worth making the switch now so we can get some testing in too.

Closes #25

### DIFF
--- a/THIRD-PARTY-NOTICES.md
+++ b/THIRD-PARTY-NOTICES.md
@@ -33,7 +33,7 @@ used under the terms of the following licences:
 | orc           | [orc License](https://gitlab.freedesktop.org/gstreamer/orc/blob/master/COPYING) (BSD-like)                |
 | pango         | LGPLv3                                                                                                    |
 | pixman        | MIT Licence                                                                                               |
-| zlib          | [zlib Licence](https://github.com/madler/zlib/blob/master/zlib.h)                                         |
+| zlib-ng       | [zlib Licence](https://github.com/zlib-ng/zlib-ng/blob/develop/LICENSE.md)                                |
 
 Use of libraries under the terms of the LGPLv3 is via the
 "any later version" clause of the LGPLv2 or LGPLv2.1.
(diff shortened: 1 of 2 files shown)


lovell:
I'll merge this so we can test it with libvips v8.10.6 in the expectation we'll upgrade to the supported release when available, or revert if required.

PR #98, opened 2021-03-17
Linux x64: remove async unwind tables, ~7% smaller binary size

lovell:
This change makes getting a backtrace slightly more difficult in some circumstances, or at least the output is less useful, but with a worthwhile reduction in binary size.

It makes very little difference with macOS/clang, presumably as it is able to optimise this away already, so this change is for Linux/gcc only. There is also minimal impact on cross-compiled ARM binaries, so perhaps we should exclude them too?

Before:
```
22742608 libvips-cpp.so.42.12.7
 8166054 libvips-8.10.6-linux-x64.tar.br
 9891496 libvips-8.10.6-linux-x64.tar.gz
```
After:
```
21145168 libvips-cpp.so.42.12.7
 7767290 libvips-8.10.6-linux-x64.tar.br
 9354659 libvips-8.10.6-linux-x64.tar.gz
```

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -29,9 +29,10 @@ mkdir ${TARGET}
 export FLAGS+=" -Os -fPIC"
 
 # Force "new" C++11 ABI compliance
+# Remove async exception unwind/backtrace tables
 # Allow linker to remove unused sections
 if [ "$LINUX" = true ]; then
-  export FLAGS+=" -D_GLIBCXX_USE_CXX11_ABI=1 -ffunction-sections -fdata-sections"
+  export FLAGS+=" -D_GLIBCXX_USE_CXX11_ABI=1 -fno-asynchronous-unwind-tables -ffunction-sections -fdata-sections"
 fi
 
 # Common build paths and flags


kleisauke:
Has this been tested on the sharp test suite? I remembered that `-fno-asynchronous-unwind-tables` is quite dangerous on Windows with corrupt PNG files (e.g. `test/fixtures/truncated.png`), because `longjmp` has to unwind the stack there, see:
https://github.com/mxe/mxe/commit/b42cd62e9a4a4e583be4970ebdced357d02d5a71#r34386467

I'm not sure if this is also the case on Linux, but it could be an issue with exception handling in C++.

lovell:
The sharp tests pass with this change on Linux x64 (I've not tried ARM yet) and I guess this is due to the use of libspng rather than libpng for reading PNG files.

kleisauke:
Ah yes, I forgot about libspng for a moment. But how about dependencies that still depend on libpng for reading? For example, does this work?:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
  <image width="100" height="100" xlink:href="data:image/png;base64,IMAGE_DATA"/>
</svg>
```
(where `IMAGE_DATA` is a base64-encoded truncated PNG image)

Anyway, this is not the only place where `longjmp`'s are used, see for example:
https://github.com/libvips/libvips/blob/8d6f90c060ffedfbd787d1ddf60635bfb289f0db/libvips/foreign/vips2jpeg.c#L199
https://github.com/libvips/libvips/blob/8d6f90c060ffedfbd787d1ddf60635bfb289f0db/libvips/foreign/vips2jpeg.c#L526-L529

lovell:
Great test case, thank you. I tried it with and without the change in this PR and only the first few lines of the image are rendered, as expected, so I've added it to sharp via commit https://github.com/lovell/sharp/commit/cb592ce5888edaa233ad55082c7e780f182dc792

I've done a little research into setjmp/longjmp.

> "In Microsoft C++ code on Windows, longjmp uses the same stack-unwinding semantics as exception-handling code"
> https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/longjmp?view=msvc-160

> "most implementations of longjmp leave the stack frame intact"
> https://en.wikipedia.org/wiki/Setjmp.h#Caveats_and_limitations

> "The built-in versions are used internally by GCC’s libraries to implement exception handling on some targets."
> https://gcc.gnu.org/onlinedocs/gcc/Nonlocal-Gotos.html 

> "compatible with the GCC __builtin_setjmp implementation allowing code built with the clang and GCC to interoperate."
> "Interacting with exceptions on Windows is significantly more complicated"
> https://llvm.org/docs/ExceptionHandling.html#llvm-eh-sjlj-setjmp

I think this means:

* MSVC does (partial) stack unwinding, which means anything that runs in the same process as MSVC compiled code will need the async unwind tables, which is what you discovered.
* GCC doesn't do stack unwinding, and provides fast builtins for the special case where the return value is 1, so possibly not used in [vipspng.c](https://github.com/libvips/libvips/blob/8dc2db978958bc7e3e68fe087c31c9834a1b8c9b/libvips/foreign/vipspng.c#L153) as -1 is returned.
* Clang follows GCC's approach, except on Windows.

kleisauke:
TIL, thanks for the investigation and the added test case!

kleisauke (approved):

PR #103, opened 2021-05-24
xml2: build without support for regular expressions

kleisauke:
We can safely remove the `--with-regexps` flag, now that libxml2 v2.9.12 has been released.


kleisauke:
macOS x64 failure seems to be a temporary network issue, let me see if a rebase fixes that.

lovell:
Thank you Kleis.

PR #104, opened 2021-06-03
Linux x64: upgrade to gcc 10

lovell:
Reduces final binary size by ~2% (from 20974256 to 20539888 bytes), which I guess may be due to the following change:

> `-finline-functions` is now enabled at `-O2` and was retuned for better code size versus runtime performance trade-offs

https://gcc.gnu.org/gcc-10/changes.html

I think this is a relatively safe upgrade, but doing so via a PR for comments, if any.


kleisauke (approved):
LGTM!

PR #105, opened 2021-06-29
Optimise for binary size: aom, orc, webp

lovell:
These dependencies provide their own hand-written SIMD intrinsics and use runtime CPU detection, so compile time optimisations are much less relevant.

This might be of possible interest to @kleisauke as it reduces the binary size by ~6% without any noticeable performance drop, hence the PR.

Before:
```
 6869110 libaom.a
 1708024 liborc-0.4.a
20095232 libvips-cpp.so.42.13.0
 1120142 libwebp.a
```
After:
```
 6320086 libaom.a
 1295216 liborc-0.4.a
18751936 libvips-cpp.so.42.13.0
  779230 libwebp.a
```

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -251,7 +251,7 @@ cd ${DEPS}/aom
 mkdir aom_build
 cd aom_build
 AOM_AS_FLAGS="${FLAGS}" LDFLAGS=${LDFLAGS/\$/} cmake -G"Unix Makefiles" \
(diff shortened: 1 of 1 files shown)


kleisauke (approved):
LGTM! Do you think MozJPEG could also benefit from this?

lovell:
Yes, we should do the same for mozjpeg, good idea.

lovell:
mozjpeg updated via commit https://github.com/lovell/sharp-libvips/commit/f0b4132cae04025687f66ad8c6c4c7c38c3ac1b1

PR #106, opened 2021-07-03
glib: remove deprecated/unused GRegex via patch

lovell:
The glib maintainers started deprecating the outdated, PCRE1-based GRegex about 6 years ago:

https://bugzilla.gnome.org/show_bug.cgi?id=755693
https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1451

This means we can pre-emptively remove it as there is no dependence on it from any of the libraries we rely on.

The sharp unit tests pass using the output.

The resultant binaries are only about 150KB smaller, but it all helps.

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -208,6 +208,7 @@ cd ${DEPS}/glib
 if [ "${PLATFORM%-*}" == "linuxmusl" ] || [ "$DARWIN" = true ]; then
   $CURL https://gist.github.com/kleisauke/f6dcbf02a9aa43fd582272c3d815e7a8/raw/13049037ce45592d0eb76a12a761212bc48bc879/glib-proxy-libintl.patch | patch -p1
 fi
+$CURL https://gist.githubusercontent.com/lovell/7e0ce65249b951d5be400fb275de3924/raw/1a833ef4263271d299587524198b024eb5cc4f34/glib-without-gregex.patch | patch -p1
 LDFLAGS=${LDFLAGS/\$/} meson setup _build --default-library=static --buildtype=release --strip --prefix=${TARGET} ${MESON} \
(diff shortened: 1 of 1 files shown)


lovell:
@kleisauke Shall I submit a PR to the build-win64-mxe repo that applies the same patch? (It is possible that the "all" variant might include a dependency that does require GRegex, but I've not checked yet.)

kleisauke:
Sure, that'd be great! I had a quick look and couldn't find any occurrence of GRegex in the "all" variant.

> The resultant binaries are only about 150KB smaller, but it all helps.

A while back I investigated shrinking libgio (for wasm-vips), and came up with this patch:
https://gist.github.com/kleisauke/c680ad19d26e510c8f9ac23c6ac308d2

This seems to reduce the size a lot, but it still feels a bit risky.

Before:
```
-rw-r--r--. 1 root root 5553194 Jul  5 10:40 libgio-2.0.a
```

After:
```
-rw-r--r--. 1 root root 1108752 Jul  5 10:43 libgio-2.0.a
```

(A quick check also reveals that it fails to compile gdk-pixbuf with this patch enabled, so there's still some work to be done)

kleisauke (approved):

lovell:
Thanks Kleis, I experimented with further patching of libgio to also remove some GApplication logic, but that only made a tiny difference. Your approach is much more extreme!

kleisauke:
The libgio patch didn't help a lot after further inspection, `libvips-cpp.so.42.13.0` was shrunk by 894944 bytes (from 18444704 to 17549760 bytes, ~4.9% decrease), see commit https://github.com/kleisauke/libvips-packaging/commit/3f2110bca3d775898490e2a26c1c52a4327798ac.

I think the link-time garbage-collection flags have already done a great job of reducing the size of the library (+ this patch seems too risky).

Issue #109, opened 2021-08-03
librsvg: rustc 1.54.0 build fails on ARM64 (Linux and macOS)


lovell:
rustc 1.53.0 works but 1.54.0 fails; needs investigating.

macOS
```
stable-x86_64-apple-darwin installed - rustc 1.54.0 (a178d0322 2021-07-26)
```
```
libtool:   error: object name conflicts in archive: .libs/librsvg-2.lax/librsvg_c_api.a//Users/runner/work/sharp-libvips/sharp-libvips/deps/svg/./.libs/librsvg_c_api.a
```
Linux glibc
```
stable-x86_64-unknown-linux-gnu installed - rustc 1.54.0 (a178d0322 2021-07-26)
```
```
libtool:   error: object name conflicts in archive: .libs/librsvg-2.lax/librsvg_c_api.a//deps/svg/./.libs/librsvg_c_api.a
```

Interestingly, Linux musl ARM64 works and that uses rustc nightly so I guess whatever causes this might already be fixed there.

kleisauke:
It looks like this is the same issue as the one mentioned in https://github.com/lovell/sharp-libvips/pull/73#issuecomment-783403636. I think it's due to commit https://github.com/rust-lang/compiler-builtins/commit/96bc62e669403a82a1becfc7863bdc17f7317d68 which causes `fp_mode.o` to also conflict for non-musl ARM64 targets.

This can be resolved by rebuilding Rust's standard library using the `build-std` feature, similar as the musl ARM64 target (see e.g. commit https://github.com/lovell/sharp-libvips/pull/90/commits/9f54bfdc8dec61aadc457560a98a8e6d54ea1375). Unfortunately, this requires a nightly version of Rust.

Perhaps there's a better way to resolve this?

lovell:
I was kind of hoping this might be a new problem :sweat_smile:

Switching to nightly rustc everywhere is probably the best approach right now, plus might unlock useful new features e.g. `panic_immediate_abort` can reduce binary size.

kleisauke:
> I think it's due to commit [rust-lang/compiler-builtins@96bc62e](https://github.com/rust-lang/compiler-builtins/commit/96bc62e669403a82a1becfc7863bdc17f7317d68) which causes `fp_mode.o` to also conflict for non-musl ARM64 targets.

[...]

PR #110, opened 2021-08-07
Use Rust's build-std feature for all ARM64 targets

kleisauke:
Resolves: #109.


lovell:
Brilliant, thank you.

PR #111, opened 2021-08-07
Bump dep: glib

kleisauke:
(I've no idea why Anitya sees this as pre-release: https://release-monitoring.org/project/10024/)

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -99,7 +99,7 @@ CURL="curl --silent --location --retry 3 --retry-max-time 30"
 # Dependency version numbers
 VERSION_ZLIB_NG=2.0.5
 VERSION_FFI=3.4.2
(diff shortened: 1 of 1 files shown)


lovell:
Most Gnome projects use an odd/even system in the minor part of semver to represent "development" vs "stable" respectively. The stability refers to the API rather than code quality, so we can happily use the latest version regardless. Maybe we should change the `stable_versions` check for Gnome projects?

kleisauke:
TIL, thanks for the info! We could use Anitya's `latest_version` JSON field for GNOME projects, but that field also tracks actual pre-release versions. Perhaps a 4th parameter called `xfail` in the `version_latest` function that doesn't set `ALL_AT_VERSION_LATEST` to `false` would be simpler.

PR #123, opened 2021-12-21
ARM64: build librsvg without the -Zbuild-std feature flag

kleisauke:
This feature flag is no longer necessary after PR rust-lang/compiler-builtins#444. 

The only downside I can think of is that the standard library is now built without `panic_immediate_abort`.

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -184,14 +184,7 @@ if [ "$DARWIN" = true ]; then
   curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
     | sh -s -- -y --no-modify-path --profile minimal ${DARWIN_ARM:+--default-toolchain nightly}
   if [ "$DARWIN_ARM" = true ]; then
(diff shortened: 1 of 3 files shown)


lovell:
Thanks Kleis, this definitely simplifies things and speeds up the ARM64 builds by about 10%.

The final binary size is a little bigger, as expected, perhaps ~4%. There's a big difference in the size of the intermediate librsvg static library, which highlights how much extra code the `#[derive(Debug)]` trait can result in.

Before:
```
-rw-r--r--    1 root     root       8942772 Dec 20 21:53 librsvg-2.a
-rwxr-xr-x    1 root     root      17326880 Dec 20 21:56 libvips-cpp.so.42.14.1
```
After:
```
-rw-r--r--    1 root     root      12015254 Dec 21 10:21 librsvg-2.a
-rwxr-xr-x    1 root     root      17953656 Dec 21 10:24 libvips-cpp.so.42.14.1
```

PR #126, opened 2022-01-30
Bump deps: fontconfig (switch to meson, patch known leaks)

lovell:
Adds 4 patches made to fontconfig's `main` branch since v2.13.94.

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -115,7 +115,7 @@ VERSION_PROXY_LIBINTL=0.3
 VERSION_GDKPIXBUF=2.42.6
 VERSION_FREETYPE=2.11.1
 VERSION_EXPAT=2.4.4
-VERSION_FONTCONFIG=2.13.93
+VERSION_FONTCONFIG=2.13.94
 VERSION_HARFBUZZ=3.2.0
 VERSION_PIXMAN=0.40.0
 VERSION_CAIRO=1.17.4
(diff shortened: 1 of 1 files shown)


kleisauke (approved):
LGTM!

lovell:
Thanks, it's removed most of the memory leaks I was previously seeing, but we're still impacted by what I think is caused by https://gitlab.freedesktop.org/fontconfig/fontconfig/-/issues/302

```
==102705== 321 bytes in 32 blocks are indirectly lost in loss record 8,127 of 8,480
==102705==    at 0x52A5045: malloc (vg_replace_malloc.c:381)
==102705==    by 0x573A50E: strdup (strdup.c:42)
==102705==    by 0x12B3EDFC: FcValueSave
==102705== 
```

Rather luckily this has just been fixed upstream so I'll add this patch too.

https://gitlab.freedesktop.org/fontconfig/fontconfig/-/merge_requests/220

kleisauke:
If preferred, I could make a v8.12.2-build2 release at https://github.com/libvips/build-win64-mxe with the same patches incorporated (+ glib, expat and lcms2 version bumps).

lovell:
Yes please, that would be great.

kleisauke:
https://github.com/libvips/build-win64-mxe/releases/tag/v8.12.2-build2 :tada:

PR #130, opened 2022-03-28
Add linux-s390x as a target

nealef:
There is no CentOS 7 image for Linux on z (aka s390x) so we use the ClefOS image which is a CentOS clone built for s390x. This is an official image of Docker. Other changes include parameterizing the pulling of docker images so that if a target platform is specified only that image required for the build is pulled.

* build.sh
  - Add linux-s390x
  - Parameterize the docker images
  - Option to pull only selected images

* build/lin.sh
  - Fix URLs
  - vips needs to run autogen.sh to generate configure

* linux-s390x
  - Dockerfile: Build using ClefOS:7 (an official Docker image for s390x)
  - meson.ini: Copy of x64
  - Toolchain.cmake: Copy of x64


lovell:
Hello, thanks for the PR. Is there a CI provider that offers native support for this architecture to open source projects (we currently use GitHub Actions for amd64 and CircleCI for aarch64)?

nealef:
What do you need? I have a Jenkins server available and can install something else if it's available.

lovell:
Thanks, I'm looking for ways to automate the following (in priority order), if possible as part of this PR.

* Ensure all future updates to this repo and the dependencies will continue to compile without failure on s390x
* Build and publish s390x tarballs as part of a [tagged release](https://github.com/lovell/sharp-libvips/releases)

nealef:
What do you use for your other targets?

build.sh
```diff
@@ -87,11 +104,11 @@ for flavour in win32-ia32 win32-x64 win32-arm64v8; do
...
 
-# Linux (x64, ARMv6, ARMv7, ARM64v8)
-for flavour in linux-x64 linuxmusl-x64 linux-armv6 linux-armv7 linux-arm64v8 linuxmusl-arm64v8; do
+# Linux (x64, ARMv6, ARMv7, ARM64v8, s390x)
+for flavour in linux-x64 linuxmusl-x64 linux-armv6 linux-armv7 linux-arm64v8 linuxmusl-arm64v8 linux-s390x; do
   if [ $PLATFORM = "all" ] || [ $PLATFORM = $flavour ]; then
     echo "Building $flavour..."
     docker build -t vips-dev-$flavour $flavour
-    docker run --rm -e "VERSION_VIPS=${VERSION_VIPS}" -e VERSION_LATEST_REQUIRED -v $PWD:/packaging vips-dev-$flavour sh -c "/packaging/build/lin.sh"
+    docker run -it --rm -e "VERSION_VIPS=${VERSION_VIPS}" -e VERSION_LATEST_REQUIRED=false -v $PWD:/packaging:z vips-dev-$flavour sh -c "/packaging/build/lin.sh"
```
lovell, line 112: The `VERSION_LATEST_REQUIRED` env var should be passed through as-is.

build/lin.sh
```diff
@@ -455,21 +455,22 @@ fi
...
+$CURL https://github.com/dloebl/cgif/archive/refs/tags/V${VERSION_CGIF}.tar.gz | tar xzC ${DEPS}/cgif --strip-components=1
 cd ${DEPS}/cgif
 CFLAGS="${CFLAGS} -O3" meson setup _build --default-library=static --buildtype=release --strip --prefix=${TARGET} ${MESON} \
   -Dtests=false
 ninja -C _build
 ninja -C _build install
 
 mkdir ${DEPS}/vips
-$CURL https://github.com/libvips/libvips/releases/download/v${VERSION_VIPS}/vips-${VERSION_VIPS}.tar.gz | tar xzC ${DEPS}/vips --strip-components=1
+$CURL https://github.com/libvips/libvips/archive/refs/tags/${VERSION_VIPS}.tar.gz | tar xzC ${DEPS}/vips --strip-components=1
```
lovell, line 466: Is there something we can change/fix in the tarballs that libvips publishes to improve s390x support and avoid running autogen? (We're switching libvips to meson, so this problem might go away.)

nealef: I don't think this is s390x specific. The tarball that is downloaded doesn't have configure built so you have to use autogen.sh to create it. The original URL wasn't finding the source tarball anymore.

build.sh
nealef: D'oh! I missed removing this.

build/lin.sh
nealef: That problem was caused by me specifying v8.12.2 as an argument and not just 8.12.2. I'll push my stuff shortly. I just need to work out how I can integrate the s390x build/builder into the scheme.

nealef:
Starting afresh. Closing this one.

PR #131, opened 2022-03-30
Add linux-s390x as a target

nealef:
* build.sh 
  - Add linux-s390x
   - Parameterize the docker images
   - Option to pull only selected images

* build/lin.sh
   - Fix URL

* linux-s390x
  - Dockerfile: Build using ClefOS:7 (an official Docker image for s390x)
  - meson.ini: Copy of x64
  - Toolchain.cmake: Copy of x64


lovell:
Thanks for the updates, did you have any thoughts or suggestions for the CI process I mentioned in https://github.com/lovell/sharp-libvips/pull/130#issuecomment-1080469081 ?

> * Ensure all future updates to this repo and the dependencies will continue to compile without failure on s390x
> * Build and publish s390x tarballs as part of a [tagged release](https://github.com/lovell/sharp-libvips/releases)

nealef:
I'm looking at various CI options. As I mentioned I have `jenkins` available and I have just installed and configured `gitlab-runner` to see what it can do but it's a learning curve.

nealef:
I am still working on this but have been distracted by travel etc.

lovell:
Thanks for the update Neale, no worries and no hurry.

If you hadn't already seen it, I wonder how slow QEMU emulation of s390x might be via something like https://github.com/marketplace/actions/run-on-architecture

nealef:
> Thanks for the update Neale, no worries and no hurry.

I have a github action runner that should be able to listen for work from your git repo. I just need a token.

piggynl:
Hi, is there any update on this? May I help you if you need it? I'm wondering if we can add linux-riscv64 as another target and do that based on the same changes on the CI process.

nealef:
I had attempted to set up a server that could be used by your CI/CD infrastructure but because I don't own the domain I can't get certificates etc. I am researching whether the folks running the Linux Community System (or whatever they call it now) could make server(s) available.

seenusuvarna:
Any updates on getting this into the baseline?

nealef:
> Any updates on getting this into the baseline?

It's a matter of trying to source a server to run CI upon for s390x. I have several servers but I don't own the domain so I can't get certificates to set up github actions. I am looking to see if I can get CircleCI on trial but may end up with the same problem.

seenusuvarna:
This might be a good resource to use:
https://www.ibm.com/community/z/open-source/virtual-machines-request/
Or there are LinuxOne resources on IBM Cloud that can be used that are free.

nealef:
My systems are all on that facility. I have the VMs to be able to do the CI work but not the ownership of the domain to get certificates for them.

nealef:
CircleCI now supports s390x. I have installed the agent on a VM running the equivalent of CentOS 7.9. I will also create one for RHEL8. I need to complete the `launch-agent-config.yaml` configuration with an `AUTH_TOKEN` is that something that gets created at your end? I have gone to the [setup page](https://app.circleci.com/projects/setup/) but am not sure what I need to provide (i.e. the location of the `launch.yaml`). I am using my github id as the `organization name`. So I am not proceeding until I know what's required.

lovell:
Thanks for all your work on this Neale. I've taken a slightly different approach with commit https://github.com/lovell/sharp-libvips/commit/80b01e3926b93d060e6112a6b0076b5e5f1ada2d and the use of cross-compilation, which makes it a little easier to integrate with the current CI setup.

https://github.com/lovell/sharp-libvips/actions/runs/6364941981/job/17281581537

I also plan to start providing prebuilt *sharp* binaries for s390x, and these will be built and tested on CI via the QEMU-based https://github.com/uraimo/run-on-arch-action (matches the approach for 32-bit ARMv6).

These should be available from the forthcoming sharp v0.33.0 onwards, enjoy!

PR #133, opened 2022-05-13
Use `git ls-remote` to fetch the latest tag

kleisauke:
Using the `GITHUB_TOKEN` secret didn't work, since that access token
is scoped to the repository running the workflow. So, retrieving tags
from other repositories might still exceed the rate limit of GitHub's
Rest API.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -49,8 +49,6 @@ jobs:
     - name: Build ${{ matrix.platform }}
       id: build-release
       run: ./build.sh $(cat LIBVIPS_VERSION) ${{ matrix.platform }}
-      env:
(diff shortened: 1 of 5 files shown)


kleisauke:
Commit 925b78586b842d4d60c37d026d706a427f40d141 was also needed on Alpine-based containers to ensure the correct version sort order:
```bash
$ docker run -it alpine:3.11 sh -c "printf '2.0.3\nv4.0.3\n' | sort -V"
v4.0.3
2.0.3
$ docker run -it alpine:3.11 sh -c "\
    apk add coreutils && \
    printf '2.0.3\nv4.0.3\n' | sort -V"
2.0.3
v4.0.3
```

Otherwise, it would fail with:
```
mozjpeg version 4.0.3 has been superseded by 2.0.3
```

lovell:
This is a much better approach, thanks Kleis.

PR #136, opened 2022-05-30
arm64v8: prefer latest stable version of Rust

kleisauke:
The underlying issue (#109) should be fixed with compiler-builtins
v0.1.55, available in Rust 1.59.0.


lovell:
Thank you for all these PRs Kleis, I'll merge them in the order opened.

PR #137, opened 2022-05-30
linux-armv{7,6}: remove `--allow-multiple-definition` linker flag

kleisauke:
This should be fixed with compiler-builtins v0.1.69, available in
Rust 1.60.0.


lovell:
Nice, this should pave the way for more Rust-based dependencies too.

PR #139, opened 2022-05-30
heif: update patch URLs

kleisauke:
This is a non-functional change except for the latest patch URL, as that has been replaced by commit https://github.com/kleisauke/libheif/commit/0d44224914946a00d293c08bbaf4553acc985802 (which is a combination of cherry-picked commits available upstream).

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -273,15 +273,16 @@ mkdir ${DEPS}/heif
 $CURL https://github.com/strukturag/libheif/releases/download/v${VERSION_HEIF}/libheif-${VERSION_HEIF}.tar.gz | tar xzC ${DEPS}/heif --strip-components=1
 cd ${DEPS}/heif
 # [PATCH] aom encoder: improve performance by ~2x using new 'all intra'
-$CURL https://github.com/lovell/libheif/commit/de0c159a60c2c50931321f06e36a3b6640c5c807.patch | patch -p1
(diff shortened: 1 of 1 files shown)


kleisauke:
Probably https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43073 would be fixed if libvips adds 
```c
		options->strict_decoding = TRUE;
```
after line:
https://github.com/libvips/libvips/blob/6260a37136973f3d231d0e2c28af6190126c664e/libvips/foreign/heifload.c#L940
(untested and requires an unreleased version of libheif)

Note that the above patch doesn't implement the strict decoding modes (and the `HEIF_WARN_OR_FAIL` macro). So, for example this: 
https://github.com/strukturag/libheif/blob/37c5ae0d6d4b90011f2e4b25a06e93b4e29ff4b0/libheif/heif_decoder_aom.cc#L216-L218

Is implemented as:
https://github.com/kleisauke/libheif/blob/0d44224914946a00d293c08bbaf4553acc985802/libheif/heif_decoder_aom.cc#L203-L205
(since I wasn't sure if we should make that strict by default)

lovell:
Thank you, this will tide us over until the next libheif release.

There's also a new, as-yet-unreleased `heif_image_get_decoding_warnings` API that we might want to integrate with if we're also going to set `strict_decoding`.

PR #140, opened 2022-05-30
Cleanup configurations

kleisauke:
- Use `-target arm64-apple-macos11` when cross-compiling for macOS ARM64.
- Avoid specifying binaries with their full path.


lovell:
Much cleaner, dankjewel.

PR #142, opened 2022-06-22
glib: update patch

kleisauke:
This patch needed to be updated after https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2734.

It seems that it is now only needed for macOS, so remove the `linuxmusl` condition.


lovell:
Thank you for fixing this, looks like the latest change worked, shall I merge?

kleisauke:
Oh wait, hold on. 
```
libintl.so.8 does not exist in /target/lib
```

kleisauke:
OK, it should be fine now.

lovell:
Brilliant, thank you.

PR #143, opened 2022-06-22
macOS: remove redundant system libraries link

kleisauke:
This became redundant after https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2413.


lovell:
Dankjewel!

PR #150, opened 2022-07-22
Upgrade Python in CentOS 7 containers

kleisauke:
Python 3.6 is EOL and Meson >= 0.62.0 requires Python 3.7 or newer. See:
https://mesonbuild.com/Release-notes-for-0-62-0.html#minimum-required-python-version-updated-to-37

Also, prefer to install Ninja via pip.


lovell:
Brilliant, thanks Kleis.

PR #155, opened 2022-08-15
Linux x64: align with v2 standard microarchitecture

### LINKED ISSUE #153: Compile Linux x64 binaries with `-march=x86-64-v2`?
Instead of compiling the Linux x64 binaries with [ `-march=westmere`](https://github.com/lovell/sharp-libvips/blob/a40f861fae54f23d8ea396a483a2d2b93288fc91/linux-x64/Dockerfile#L39) perhaps we should lower it to `-march=x86-64-v2`? This would allow the binaries to run on Intel Nehalem, which doesn't support the AES and PCLMUL instruction sets.

Context: https://github.com/kleisauke/net-vips/issues/176.

lovell:
The use of `-march=x86-64-v2` requires gcc 11, so this uses the almost-equivalent `-march=nehalem` with gcc 9 on Alpine.

I checked the zlib-ng source code and it compiles and includes functions for all common x64 intrinsics and uses runtime-dispatch, so will still take advantage of PCMUL intrinsics on newer hardware.

There appears to be a small ~4Kb increase in binary size due to this change.

Before:
```
17104592 libvips-cpp.so.42.15.0
```
After:
```
17108688 libvips-cpp.so.42.15.0
```

Fixes #153

### DIFF
--- a/linux-x64/Dockerfile
+++ b/linux-x64/Dockerfile
@@ -36,7 +36,7 @@ RUN \
 ENV \
   PKG_CONFIG="pkg-config --static" \
   PLATFORM="linux-x64" \
-  FLAGS="-march=westmere" \
+  FLAGS="-march=x86-64-v2 -mtune=nehalem" \
   MESON="--cross-file=/root/meson.ini"
 
 COPY Toolchain.cmake /root/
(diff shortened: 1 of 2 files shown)


kleisauke (approved):
LGTM!

PR #156, opened 2022-09-04
CI: upgrade Ubuntu runner and checkout action

kleisauke:
See the individual commits for more details. Diff is best viewed with the "Hide whitespace" option due to commit f976680d770b6d1b7a5ac80b46a3beb07164f823.


lovell:
Thanks again.

PR #157, opened 2022-09-04
win32: upgrade Docker container to Debian 11 (bullseye)

kleisauke:
This container is used only for post-processing Windows binaries,
so we can safely bump to Debian 11 (bullseye).

The Linux ARMv7-A and ARMv6 containers still require Debian 10
(buster) to avoid a glibc version bump in the resulting binaries.


lovell:
Thanks Kleis.

PR #158, opened 2022-09-20
Meson: install only a subset of the installable files

kleisauke:
Requires at least Meson 0.64.0.

See https://mesonbuild.com/Installing.html#installation-tags.

Marked this PR as draft due to these TODO items:
- [x] GLib requires MR https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2905 to ensure `glibconfig.h` is tagged as `devel`.
- [x] Meson requires a new release that includes:
  - [x] PR https://github.com/mesonbuild/meson/pull/10826 - to ensure `gnome.mkenums` installed files are tagged as `devel`.
  - [x] PR https://github.com/mesonbuild/meson/pull/10959 - to ensure `custom_target` tags are automatically guessed.
- [x] ~FreeType needs to bump the minimum required Meson version to 0.60 and tag [`ftoption.h`](https://gitlab.freedesktop.org/freetype/freetype/-/blob/33ceac2afc3608be2d12b3dece071a91e34c3062/meson.build#L356) as `devel`.~
  No longer needed with Meson 0.64.0.
- [x] ~FriBidi needs to bump the minimum required Meson version to 0.60 and tag [`fribidi-unicode-version.h`](https://github.com/fribidi/fribidi/blob/3585a81e625e5b35962bed11fdc99d54485048e0/gen.tab/meson.build#L39) as `devel`.~
  No longer needed with Meson 0.64.0.

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -188,8 +188,7 @@ if [ "${PLATFORM%-*}" == "linuxmusl" ] || [ "$DARWIN" = true ]; then
   $CURL https://github.com/frida/proxy-libintl/archive/${VERSION_PROXY_LIBINTL}.tar.gz | tar xzC ${DEPS}/proxy-libintl --strip-components=1
   cd ${DEPS}/proxy-libintl
   meson setup _build --default-library=static --buildtype=release --strip --prefix=${TARGET} ${MESON}
-  ninja -C _build
-  ninja -C _build install
+  meson install -C _build --tag devel
 fi
 
 mkdir ${DEPS}/zlib-ng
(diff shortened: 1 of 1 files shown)


kleisauke:
This is currently blocked on PR https://github.com/lovell/sharp-libvips/pull/167, which ensures that we always use the latest Meson version, even on macOS.

Note that this PR only avoids Meson having to install things in the `bin/` and `share/` directories or that we accidentally build a shared library and depend on that, so I'm not sure if this was worth it (however, this found a couple of upstream bugs, which are now fixed).

For now, I confirmed that there are no changes to the `lib/` and `include/` directories on Linux x64 (i.e., the contents of the created tarballs are identical to those before this PR). :smile:

kleisauke:
> This is currently blocked on PR #167, which ensures that we always use the latest Meson version, even on macOS.

No longer a blocker, this PR may land independently. This is ready for review now.

lovell:
Thanks, this is a nice simplification, plus it's always good to fix upstream bugs in Meson.

PR #163, opened 2022-11-03
macOS: install Meson and Ninja via Homebrew

kleisauke:
Since GitHub actions migrated the Python installation to a universal package, which no longer adds the pip3 site-packages to the `PATH`, see: https://github.com/actions/runner-images/issues/6496.


lovell:
Thanks Kleis!

PR #164, opened 2022-11-08
macOS: upgrade earliest supported version to 10.13

lovell:
The documentation already says `>= 10.13`

https://sharp.pixelplumbing.com/install#prebuilt-binaries

Also:

* Block use of API newer than deployment target, will fail compilation if found
* Temporarily patch orc to prevent use of `pthread_jit_` functions on 10.x (x64)


kleisauke (approved):
Great! I'll see if I can resolve this in Meson as well, which would remove the need for this sed replacement and possibly MR https://gitlab.freedesktop.org/gstreamer/orc/-/merge_requests/78.

lovell:
Looks like the latest meson v0.64.0 as installed via Homebrew is broken (updated < 1 hour ago).

kleisauke:
Ouch, perhaps removing `brew update` would temporally fix that?
https://github.com/lovell/sharp-libvips/blob/34c25b5b93417d67d408d3385abaeb1832027e5b/.github/workflows/build-release.yml#L44

IIUC, that would prevent Meson from being updated. But when GitHub updates its macOS runners it probably starts failing again. :(

kleisauke:
... looking at the logs, I see this:
```
==> Pouring python@3.11--3.11.0.big_sur.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3-3.11
Target /usr/local/bin/2to3-3.11
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3-3.11'

To force the link and overwrite all conflicting files:
  brew link --overwrite python@3.11

To list all files that would be deleted:
  brew link --overwrite --dry-run python@3.11

Possible conflicting files are:
/usr/local/bin/2to3-3.11 -> /Library/Frameworks/Python.framework/Versions/3.11/bin/2to3-3.11
/usr/local/bin/idle3.11 -> /Library/Frameworks/Python.framework/Versions/3.11/bin/idle3.11
/usr/local/bin/pydoc3.11 -> /Library/Frameworks/Python.framework/Versions/3.11/bin/pydoc3.11
/usr/local/bin/python3.11 -> /Library/Frameworks/Python.framework/Versions/3.11/bin/python3.11
/usr/local/bin/python3.11-config -> /Library/Frameworks/Python.framework/Versions/3.11/bin/python3.11-config
````

Looks like Homebrew is migrating to Python 3.11.
https://github.com/Homebrew/homebrew-core/issues?q=python@3.11

We could consider reverting https://github.com/lovell/sharp-libvips/pull/163 if this happens more often. Perhaps that issue was fixed with commit https://github.com/actions/runner-images/commit/f6632ff53b24db9705c5b15bf90dbf05cc3fe295?

**Comment by lovell** [comment]
Ah, maybe we should try using `actions/setup-python`? I'll try in a separate PR.

The sharp CI environment uses this to pin Python v3.10 as the version of `node-gyp` provided by Node.js 14 doesn't work with v3.11.

https://github.com/lovell/sharp/blob/ca22af203f77dc484d57e934e3a2c5b057e095f6/.github/workflows/ci.yml#L81-L85

**Comment by lovell** [comment]
Quickest solution was to temporarily remove `brew update` for now via https://github.com/lovell/sharp-libvips/commit/033f05adad92ea1572d05b7e4db8e9676a20ca23

This can be added back once every Python-dependent package in Homebrew has been updated.

**Comment by ongjiinjoo** [comment]
Does this mean that libvips older than 10.13 is no longer supported? I have a globally built libvips (vips-8.11.4-Thu Sep 23 09:40:01 UTC 2021) but no matter what I do, npm install --build-from-source still tries to download the latest 8.13.3, which results in the same error:

```
dyld: lazy symbol binding failed: Symbol not found: _pthread_jit_write_protect_supported_np
  Referenced from: /Users/jiinjoo/repos/media-service/node_modules/sharp/build/Release/../.././vendor/8.13.3/darwin-x64/lib/libvips-cpp.42.dylib
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _pthread_jit_write_protect_supported_np
  Referenced from: /Users/jiinjoo/repos/media-service/node_modules/sharp/build/Release/../.././vendor/8.13.3/darwin-x64/lib/libvips-cpp.42.dylib
  Expected in: /usr/lib/libSystem.B.dylib
```

lovell:
@ongjiinjoo Please see https://github.com/lovell/sharp/issues/3438 for the details behind this.

PR #165, opened 2022-11-10
glib: remove slice allocator (via Alpine Linux patch)

lovell:
This patch, from Alpine Linux, is approximately equivalent to setting `G_SLICE=always-malloc` and goes further by removing all the unused code. This should help prevent segfaults/hangs on QEMU-based emulated/interpreted environments, amongst others.

(The default memory allocator included in Windows UCRT is, I think, still bad enough to warrant retaining glib's slice logic in the Windows binaries.)

Related:

https://github.com/lovell/sharp/issues/3082
https://github.com/lovell/sharp/issues/3448

https://gitlab.gnome.org/GNOME/glib/-/issues/1079
https://gitlab.com/qemu-project/qemu/-/issues/285

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -213,6 +213,10 @@ if [ "$DARWIN" = true ]; then
   $CURL https://gist.github.com/kleisauke/f6dcbf02a9aa43fd582272c3d815e7a8/raw/75b1e06250bdb0df067be4a5db54df960f35c46d/glib-proxy-libintl.patch | patch -p1
 fi
 $CURL https://gist.github.com/kleisauke/284d685efa00908da99ea6afbaaf39ae/raw/ce7f85f337357555c3112b36d4b2753ae996f5ff/glib-without-gregex.patch | patch -p1
+if [ "$LINUX" = true ]; then
+  # use malloc rather than slice allocator https://gitlab.gnome.org/GNOME/glib/-/issues/1079
+  $CURL https://raw.githubusercontent.com/alpinelinux/aports/ba57357f7a7094d1ebf4542912661383bafcf348/main/glib/always-malloc.patch | patch -p1
+fi
 meson setup _build --default-library=static --buildtype=release --strip --prefix=${TARGET} ${MESON} \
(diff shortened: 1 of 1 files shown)


kleisauke:
It looks like this patch has been discussed upstream at https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2935, but refused as it was "too aggressive". Also, the GLib maintainers still think that using the slice allocator is still necessary on macOS and Windows.

Given this, perhaps we should only apply this patch to the Linux binaries? We can also consider applying https://github.com/alpinelinux/aports/raw/ba57357f7a7094d1ebf4542912661383bafcf348/main/glib/always-malloc.patch instead, that one was mentioned in comment https://gitlab.gnome.org/GNOME/glib/-/issues/1079#note_1566854 (and could be easily `#ifdef`'d for Linux, if necessary).

BTW, the Windows benchmark mentioned in comment https://gitlab.gnome.org/GNOME/glib/-/issues/1079#note_1568456 was run in CI (not sure if that's reliable) and build against the MSVCRT variant, see:
https://gitlab.gnome.org/GNOME/glib/-/blob/main/.gitlab-ci/mingw.Dockerfile#L1
https://fedoraproject.org/wiki/Changes/F37MingwUCRT

And the discussion in:
https://news.ycombinator.com/item?id=31961912 (ignore the blog post, that was later marked as draft with commit
https://github.com/ErikMcClure/erikmcclure.github.io/commit/30407f93a0998f1b4c3419acd16f4ac9a4a1efad)

kleisauke (approved):
LGTM!

lovell:
Thanks for reviewing this Kleis, I agree we should stick to the Linux-only approach the glib maintainers are planning/considering.

kleisauke:
It looks like the GSlice allocator [will be removed in GLib 2.75.3](
https://gitlab.gnome.org/GNOME/glib/-/commit/45b5a6c1e56d5b73cc5ed798ef59a5601e56c170). :tada:

PR #166, opened 2022-11-12
Bump dep: glib

kleisauke:
- Rebase `glib-without-gregex.patch`.
- Ensure we disable the inotify backend for musl.


lovell:
Thank you!

PR #168, opened 2022-11-19
Bump deps: heif, pango

kleisauke:
- Build libheif with CMake, as preferred by upstream since v1.14.0.
- Ensure to always specify `CMAKE_BUILD_TYPE` for CMake projects.
- Prefer to specify `CMAKE_INSTALL_LIBDIR` as relative path.

---
If necessary, I could also backport commit https://github.com/strukturag/libheif/commit/ee1e5087d2cc289a161de44c013e27300224529b. 

Note that calling `heif_init()` should not be required by libvips, despite what is mentioned in the [release notes of v1.14.0](https://github.com/strukturag/libheif/releases/tag/v1.14.0) - see:
https://github.com/strukturag/libheif/blob/16385339069d4bd931efd797032039255c75dfc2/libheif/heif.h#L310-L311

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -109,10 +109,10 @@ VERSION_HARFBUZZ=5.3.1
 VERSION_PIXMAN=0.42.2
 VERSION_CAIRO=1.17.6
 VERSION_FRIBIDI=1.0.12
-VERSION_PANGO=1.50.11
+VERSION_PANGO=1.50.12
 VERSION_SVG=2.55.1
 VERSION_AOM=3.5.0
-VERSION_HEIF=1.13.0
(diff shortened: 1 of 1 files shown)


kleisauke:
This was temporarily marked as draft due to:
```
External image format libraries
...
  HEIC/AVIF load/save with libheif  : NO (dynamic module: NO)
```
Should be fixed now.

(I'm aware that such issues can be detected in Meson by passing `-Dauto_features=enabled`)

lovell:
Dankjewel! No worries about the "Defer NCLX profile..." patch as there wasn't really undefined behaviour before; I submitted it to silence UBSan so it can continue to look for real problems.

PR #170, opened 2022-12-24
tiff: avoid building unnecessary things

kleisauke:
In line with commit kleisauke/wasm-vips@628f6fa and libvips/build-win64-mxe@3c04aa7.


lovell:
Thank you Kleis, and a happy Christmas to you.

PR #171, opened 2023-01-04
Add support for OpenJPEG

steveh:
[JPEG 2000](https://en.wikipedia.org/wiki/JPEG_2000) images are widely used in the [GLAM](https://en.wikipedia.org/wiki/GLAM_(cultural_heritage)) sector, especially with [IIIF](https://en.wikipedia.org/wiki/International_Image_Interoperability_Framework).

Adding support in the prebuilt binary would be very useful!

### DIFF
--- a/build/lin.sh
+++ b/build/lin.sh
@@ -114,6 +114,7 @@ VERSION_SVG=2.55.90
 VERSION_AOM=3.5.0
 VERSION_HEIF=1.14.1
 VERSION_CGIF=0.3.0
+VERSION_OPENJPEG=2.5.0
 
 # Remove patch version component
 without_patch() {
@@ -173,6 +174,7 @@ version_latest "svg" "$VERSION_SVG" "5420" "unstable"
 version_latest "aom" "$VERSION_AOM" "17628"
 version_latest "heif" "$VERSION_HEIF" "strukturag/libheif"
 version_latest "cgif" "$VERSION_CGIF" "dloebl/cgif"
+version_latest "openjpeg" "$VERSION_OPENJPEG" "2550"
 if [ "$ALL_AT_VERSION_LATEST" = "false" ]; then exit 1; fi
 
 # Download and build dependencies from source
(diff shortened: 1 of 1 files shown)


lovell:
Hi, thanks for taking the time to open a PR.

I realise that JPEG2000 is still used in a few areas (medicine, cinemas, on the planet Mars) but its use on the web, which is the primary concern for sharp, is too low. https://caniuse.com/jpeg2000

I'm aware of (slowly) increasing adoption of JPEG-XL in the world of GLAM, and I think that format has a better chance of future possible inclusion in prebuilt binaries, especially if one of the web browsers can kick-start this.

The default behaviour for sharp is that input images will be untrusted. OpenJPEG is currently (partially) fuzz tested, but it is not integrated into libvips' [fuzz testing](https://github.com/google/oss-fuzz/tree/master/projects/libvips) build, plus I have concerns about vulnerabilities discovered by others not being released in a timely manner.

sharp will always support a globally-installed libvips, and from this PR it looks like you're comfortable with building it from source, so I recommend you take this route if you need JPEG2000 support. Make sure you trust the source of the input images.

As an aside, I couldn't get OpenJPEG to build using the change in this PR (see linker error below). You may need to add the (erroneously-named?) `-DBUILD_CODEC=FALSE` flag to prevent it compiling command line utilities.
```
[ 94%] Building C object src/bin/jp2/CMakeFiles/opj_dump.dir/converttif.c.o
/opt/rh/devtoolset-11/root/usr/libexec/gcc/x86_64-redhat-linux/11/ld: /target/lib/libtiff.a(tif_jpeg.o): in function `TIFFjpeg_create_decompress':
(.text.TIFFjpeg_create_decompress+0x11): undefined reference to `jpeg_std_error'
```

steveh:
Thanks for the considered explanation! I was able to get it to build and link with your suggestion. I also had to update the paths to cgif and vips, which is probably also necessary in `main`.

There are no user-uploaded images in our use case so I'm happy to link against openjpeg.

Would you consider a way to selectively enable building with openjpeg? i.e. with `TRUSTED=true` or the like? Just to avoid maintaining a fork. No worries if not, it's not terribly difficult to keep up.

lovell:
Glad you got it working. I'm going to close this as I don't intend to support OpenJPEG with the prebuilt binaries.

I would be happy to provide a list (in the readme?) of well-maintained forks that offer additional formats/libraries, so please do let me know if/when you'd want me to consider your work for this.

PR #187, opened 2023-07-20
archive: disable build of bsdunzip utility

kleisauke:
In line with the Windows builds, where this isn't built by default.
https://github.com/libarchive/libarchive/blob/bd074c2531e867078788fe8539376c31119e4e55/configure.ac#L277-L280

Should fix the CI for the `linuxmusl-*` platforms.


lovell:
Dankjewel Kleis.

PR #193, opened 2023-09-18
Ensure zlib-ng is installed in `lib/`

kleisauke:
Otherwise, it might inadvertently end up in the `lib64/` directory, resulting in the use of zlib 1.2.11 as provided by GLib as fallback.

<details>
  <summary>Details</summary>

```
-- Installing: /target/lib64/libz.a
-- Installing: /target/include/zlib.h
-- Installing: /target/include/zlib_name_mangling.h
-- Installing: /target/include/zconf.h
-- Installing: /target/lib64/pkgconfig/zlib.pc

...

Run-time dependency zlib found: NO (tried pkgconfig, cmake and system)
Looking for a fallback subproject for the dependency zlib
Downloading zlib source from https://zlib.net/fossils/zlib-1.2.11.tar.gz
Downloading zlib patch from https://wrapdb.mesonbuild.com/v2/zlib_1.2.11-6/get_patch

Executing subproject zlib 

zlib| Project name: zlib
zlib| Project version: 1.2.11
zlib| C compiler for the host machine: cc (gcc 11.2.1 "cc (GCC) 11.2.1 20220127 (Red Hat 11.2.1-9)")
zlib| C linker for the host machine: cc ld.bfd 2.36.1-1
zlib| C compiler for the build machine: cc (gcc 11.2.1 "cc (GCC) 11.2.1 20220127 (Red Hat 11.2.1-9)")
zlib| C linker for the build machine: cc ld.bfd 2.36.1-1
zlib| Compiler for C supports arguments -Wno-implicit-fallthrough: YES
zlib| Compiler for C supports arguments -Wno-implicit-function-declaration: YES
zlib| Build targets in project: 1
zlib| Subproject zlib finished.

Dependency zlib from subproject subprojects/zlib-1.2.11 found: YES 1.2.11
```
</details>

This issue was known to affect the linux-x64 and linuxmusl-x64 targets.

Use the built-in `wrap-mode=nofallback` option in Meson to prevent any further regressions.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -54,12 +54,12 @@ for flavour in darwin-x64 darwin-arm64v8; do
     export FLAGS="-fno-stack-check"
     # Prevent use of API newer than the deployment target
     export FLAGS+=" -Werror=unguarded-availability-new"
+    export MESON="--cross-file=$PWD/$PLATFORM/meson.ini"
 
     if [ $PLATFORM = "darwin-arm64v8" ]; then
       # ARM64 builds work via cross compilation from an x86_64 machine
       export CHOST="aarch64-apple-darwin"
       export FLAGS+=" -target arm64-apple-macos11"
-      export MESON="--cross-file=$PWD/$PLATFORM/meson.ini"
       # macOS 11 Big Sur is the first version to support ARM-based macs
       export MACOSX_DEPLOYMENT_TARGET="11.0"
       # Set SDKROOT to the latest SDK available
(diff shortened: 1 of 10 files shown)


kleisauke:
FWIW, it looks like this was regressed since commit ade8c51dd5d8c3ca95559be7efde62d27a48c690, as zlib-ng didn't use the standard CMake `GNUInstallDirs` module in earlier versions, see: 
https://github.com/zlib-ng/zlib-ng/commit/fb4f9569c661976e2680974da3ff616e2ae0056f

PR #195, opened 2023-09-24
Bump dep: cairo

kleisauke:


lovell:
Oh, a new stable release of cairo doesn't happen very often, thank you.

PR #196, opened 2023-10-17
Bump deps: heif, mozjpeg

kleisauke:
FWIW, the libheif bump to v1.17.0 might be a bit too risky/early, see e.g.:
https://github.com/strukturag/libheif/compare/v1.17.0...master

Specifically the fix in commit https://github.com/strukturag/libheif/commit/18d4e550ad7ee114fb95ba2256ee3e756174e92d, which would otherwise prevent AVIF images with premultiplied alpha channels from being decoded.


lovell:
Yeah, there were a lot of libheif changes just before the latest version was published that are unlikely to have seen enough fuzz testing yet. Those CMake failures should go away if we switch this PR to target the `upgrade-linux-containers` branch.

kleisauke:
Re-targeted to the `upgrade-linux-containers` branch instead. Commit d51dd3b9f6c80ef5946af7aa1d4bbcfc0ae6a41b is still needed for Debian Buster (an alternative for this would be to upgrade the container to `debian:buster-backports`, which provides CMake 3.18.4).

lovell:
Thanks Kleis. I suspect there'll be a patch version of libheif along very soon.

PR #199, opened 2023-10-19
Leave tarballs as-is, remove Brotli, don't re-gzip

lovell:
These files will no longer be downloaded millions of times, so there is no need to spend CPU time maximising compression, leading to faster builds etc.

@kleisauke Any problems this might cause for https://github.com/kleisauke/libvips-packaging ? I note you don't generate the Brotli-compressed tarballs anyway.


lovell:
Thanks, for sharp these tarballs are going to be re-packaged for publishing to npm (manually for now, can automate once I'm happy with the process).

PR #200, opened 2023-10-19
hwy: ensure correct lib directory

lovell:
Before:
```
2023-10-19T15:02:51.0232087Z -- Installing: /target/lib64/pkgconfig/libhwy.pc
2023-10-19T15:09:01.8740127Z     SIMD support with highway         : NO
```
After:
```
2023-10-19T18:43:41.1109738Z -- Installing: /target/lib/pkgconfig/libhwy.pc
2023-10-19T18:49:16.1614328Z     SIMD support with highway         : YES
```
Apologies for not spotting this earlier.


kleisauke (approved):
:man_facepalming:

PR #202, opened 2023-11-01
Update libtiff licence location

kleisauke:
The old URL was giving a 404 error.

Noticed while trying to integrate this file in wasm-vips.


lovell:
Thanks Kleis.

PR #213, opened 2023-12-26
Bump deps: fontconfig, glib

kleisauke:
GLib 2.79.0 requires the `packaging` Python module, so add this dependency to the Docker images as well.
> ```md
>  * Bump Meson dependency to 1.2.0 and depend on Python `packaging` module
> ```
> _From: https://gitlab.gnome.org/GNOME/glib/-/blob/2.79.0/NEWS#L25_

While we're here, fetch the latest wasm-vips commit by default using the special `HEAD` reference. On Github, this will be the tip of the default branch.


lovell:
Dankjewel Kleis!

PR #214, opened 2023-12-27
Meson: replace deprecated "pkgconfig" entry in machine files

kleisauke:
See:
https://mesonbuild.com/Release-notes-for-1-3-0.html#machine-files-pkgconfig-field-deprecated-and-replaced-by-pkgconfig

This should fix this deprecation warning:
```
DEPRECATION: "pkgconfig" entry is deprecated and should be replaced by "pkg-config"
```


lovell:
Dank!

Issue #113, opened 2021-09-03
Compile FreeType with HarfBuzz support


kleisauke:
FreeType is currently compiled without HarfBuzz support on Linux and macOS, which can lead to incorrect hinting of glyphs from languages that require more complex text shaping.

Similar as within the Windows builds, this circular dependency can be addressed as follows:

1. Compile FreeType without HarfBuzz support (`-Dharfbuzz=disabled` or `--without-harfbuzz`).
2. Compile HarfBuzz with FreeType support.
3. Compile FreeType with HarfBuzz support (`-Dharfbuzz=enabled` or `--with-harfbuzz`).

https://fedoraproject.org/wiki/Changes/EnableHarfBuzzInFreeType might also be relevant here.

Runi-c:
This problem seems to break emoji fonts entirely in the built version of libvips on Linux and macOS, as per https://github.com/lovell/sharp/issues/3959.

I spent a day trying to fix this, but it turns out that building circular dependencies with autotools is easier said than done. Specifically, I was shocked to find that `pkg-config` actually segfaults due to stack overflow when trying to traverse the dependency tree after `freetype` is rebuilt against `harfbuzz` as suggested in this issue. 

I am not experienced enough with these arcane C build systems to figure out a workaround to this segfault - the only real lead I found was that [pkgconf](http://pkgconf.org/features.html) could maybe be used instead of `pkg-config`, as it is capable of the impressive feat of resolving circular dependencies without a stack overflow.

kleisauke:
It looks like pkg-config >= 0.28 is needed to support circular `Requires` dependencies.
https://bugs.freedesktop.org/show_bug.cgi?id=7331
https://gitlab.freedesktop.org/pkg-config/pkg-config/-/commit/6d6dd43e75e2bc82cfe6544f8631b1bef6e1cf45

Amazon Linux 2 only provides version 0.27.1 and doesn't include the newer and actively maintained pkgconf, which is now generally preferred over pkg-config.

In the meantime, this issue can be resolved with a simple sed replacement, I just opened PR #218 for this.

PR #218, opened 2024-02-11
Build FreeType against HarfBuzz

kleisauke:
Resolves: #113.


lovell:
Brilliant, thank you for working this all out.

PR #219, opened 2024-02-19
Bump deps: ffi, glib, highway

kleisauke:
Also update some comments on why s390x is currently built without Highway support.


lovell:
:tada:

PR #225, opened 2024-03-15
Upgrade to libvips v8.15.2

kleisauke:


build/lin.sh
```diff
@@ -457,7 +457,7 @@ CFLAGS="${CFLAGS} -O3" meson setup _build --default-library=static --buildtype=r
 meson install -C _build --tag devel
 
 mkdir ${DEPS}/vips
-$CURL https://github.com/libvips/libvips/releases/download/v${VERSION_VIPS}/vips-$(without_prerelease $VERSION_VIPS).tar.xz | tar xJC ${DEPS}/vips --strip-components=1
+$CURL https://github.com/libvips/libvips/releases/download/v${VERSION_VIPS}a/vips-$(without_prerelease $VERSION_VIPS).tar.xz | tar xJC ${DEPS}/vips --strip-components=1
```
lovell, line 460: Thanks for this Kleis. Should we update the release label to match the tag in the libvips repo? This should allow us to remove the `a` suffix entirely.

kleisauke: Did you mean to adjust the `LIBVIPS_VERSION` file to include the extra `a`? The binaries uploaded at https://github.com/libvips/libvips/releases/tag/v8.15.2a release are _not_ suffixed with `a`, so I thought this was easier.

kleisauke: I think it's probably too late now to remove the `a`-suffix from the libvips release, since Anitya already recorded this tag: https://release-monitoring.org/project/5097/

lovell: My comment was about the release label rather than the filename, but given the tag is out in the wild I agree it's now too late.

lovell:
Thanks very much for doing this and publishing the Windows binaries too.

PR #228, opened 2024-03-24
Build FreeType against libpng

kleisauke:
It seems colored emoji glyphs are stored as PNG images in OpenType fonts, so compile FreeType with `-Dpng=enabled` to ensure that emoji can be rendered.

Tested by replacing `node_modules/@img/sharp-libvips-linux-x64/lib/libvips-cpp.so.42` with the produced binary and by doing:
```console
$ node -e "require('sharp')({text: { text: 'sharp is awesome! 🧩 ❤️', rgba: true }}).toFile('x.png')"
```
x

Context: https://github.com/lovell/sharp/issues/3959.


lovell:
Ah this is interesting, I thought "emoji" fonts stored glyphs using `COLR` vectors, but you're quite right that the common ones such as Noto still use the `CBDT` raster format, i.e. PNG.

Will we ever be able to remove libpng? ;)

PR #231, opened 2024-04-01
vips: disable HBR support in heifsave

kleisauke:
To mirror the same patch added in https://github.com/libvips/build-win64-mxe/pull/61.


lovell:
Thank you.

PR #234, opened 2024-04-14
Add libde265 as optional dependency to support HEIC format

john-trieu-nguyen:
Thought it would be useful to others building their own binaries who also need HEIC support.

I do acknowledge the current prebuilt binaries on the @img repository intentionally don't support HEIC due to surrounding legalities so have made this feature as an optional flag.

Not sure if this still introduces potential legal issues but thought I would submit the pull request anyway.

### DIFF
--- a/THIRD-PARTY-NOTICES.md
+++ b/THIRD-PARTY-NOTICES.md
@@ -18,6 +18,7 @@ used under the terms of the following licences:
 | highway       | Apache-2.0 License, BSD 3-Clause                                                                          |
 | lcms          | MIT Licence                                                                                               |
 | libarchive    | BSD 2-Clause                                                                                              |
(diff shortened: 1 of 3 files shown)


lovell:
Thank you for taking the time to work on a PR. I will be unable to merge it as I am unwilling to maintain any optional dependencies in these scripts. Hopefully your efforts will be of use to someone else in the future.

(There's no legal issue that I'm aware of around publishing source code and scripts to _build_ HEVC-related software, the patent-claim problem arises from the distribution of binaries.)

john-trieu-nguyen:
No problem, totally understand. Yes, thought it would be helpful to others searching for HEIC in the repo.

Thanks for the amazing libraries! Was a breeze to work through :)

Issue #226, opened 2024-03-17
librsvg: latest version no longer depends on gdk-pixbuf but adds ~1MB to binary size


lovell:
I didn't notice it at the time, but as part of commit https://github.com/lovell/sharp-libvips/commit/eb6fedeea4dacbc214cf978a5ba75657295ac274 librsvg was upgraded to 2.57.91, which includes a major change to the way inline images within SVG files are handled.

https://gitlab.gnome.org/GNOME/librsvg/-/blob/main/NEWS

> Librsvg no longer uses gdk-pixbuf for loading raster images; all image loading uses Rust code now.  Older versions of librsvg would use the traditional libpng and libjpeg-turbo libraries, and gdk-pixbuf's own GIF decoder.  Starting with this version of librsvg, it only uses Rust crates for image codecs.

> ...the only supported formats for the <image> element are PNG, JPEG, GIF, WEBP, and SVG

It looks like rust-based decoders are being used instead, which adds ~1MB to binary sizes.

Before:
```
 9852590 Jan 21 11:02 librsvg-2.a
16243944 Jan 21 11:03 libvips-cpp.so.42.17.1
```
After:
```
12419718 Feb 11 17:56 librsvg-2.a
17189896 Feb 11 17:57 libvips-cpp.so.42.17.1
```

We should investigate to see what, if anything, can be reduced/removed.

Thoughts of the top of my head:

- Can we now remove gdk-pixbuf and libpng? Anything else?
- Can we specify a minimal `image` rust crate with only required decoders?
- Do we investigate a switch to resvg?

kleisauke:
Ah, bummer, I should have noticed this earlier. A short-term solution would be to downgrade librsvg to v2.57.2.

> * Can we now remove gdk-pixbuf and libpng? Anything else?

Unfortunately, gdk-pixbuf is still required by the legacy pixbuf-based C API of librsvg (`rsvg_pixbuf_*`) on which the `rsvg_handle_get_pixbuf()`/`rsvg_handle_get_pixbuf_sub()` functions depend.
https://gitlab.gnome.org/GNOME/librsvg/-/blob/a9b4bfdd3e919535c12fed3947aa746d0b5dddce/librsvg-c/Cargo.toml#L16

Perhaps it's possible to feature-gate these APIs? /cc @federicomenaquintero

lovell:
[...]

PR #236, opened 2024-04-27
Bump deps: aom, archive, fribidi, rsvg

kleisauke:
Resolves: #226.

(I also included commit 39b0143c1b8fa3bc4700c1b10a622eb3aaa8baff to align with the changes made in commit https://github.com/lovell/sharp/commit/36e60bf0402f6bb241198393d522b7ff0bdc686b)


lovell:
Dankjewel Kleis for all your upstream work on librsvg to get us here too.

The file size of the intermediate static lib is rather interesting. :monocle_face: 
```
-rw-r--r-- 1 root root 90902052 Apr 27 13:04 librsvg-2.a
```

kleisauke:
Thanks! Indeed, that huge size of intermediate static lib is curious. I suspect it's due to LTO, but I'm not sure.

PR #240, opened 2024-07-08
xml2: switch from autotools to meson

kleisauke:
Before:
```
-rw-r--r-- 1 root root   878202 Jul  8 09:21 libxml2.a
```

After
```
-rw-r--r-- 1 root root   554918 Jul  8 11:34 libxml2.a
```

Marked as draft as I didn't test this at runtime yet. I suspect the reader, writer, valid, http and zlib features of libxml2 were required by libgsf instead of librsvg.


kleisauke:
Seems to work! I tested this with:
```console
$ LD_PRELOAD=$HOME/libvips-8.15.2-linux-x64/lib/libvips-cpp.so.42 npm run test
[...]
  1199 passing (4s)
```
on sharp's test suite.

lovell:
Reduced binary size and reduced attack surface, excellent, thank you Kleis.

PR #241, opened 2024-07-08
Disable build of cruft

kleisauke:
Mirrors commit libvips/libvips@ee482ca and libvips/libvips@a9e161f.


lovell:
Excellent use of the word "cruft", thank you.

Issue #206, opened 2023-11-06
Enhancement: publish packages directly to npm, including provenance


lovell:
This will remove the use of GitHub releases (git tags will remain).

https://docs.npmjs.com/generating-provenance-statements#publishing-packages-with-provenance-via-github-actions

lovell:
Publishing directly to npm from GitHub Actions would remove the ability to use 2FA so I'm unsure we should go down this route at the moment. If GitHub Actions improves support for this, e.g. prompt for an OTP in the middle of a job, then I'll revisit.

PR #242, opened 2024-07-15
ppc64le support

sumitd2:
Tested locally.
cc: @seth-priya


platforms/linux-ppc64le/Dockerfile
```diff
@@ -0,0 +1,54 @@
+FROM debian:bullseye
+LABEL maintainer="Lovell Fuller <npm@lovell.info>"
+
+# Create Debian 10 (glibc 2.28) container suitable for cross-compiling Linux s390x binaries
```
kleisauke, line 4:
```suggestion
# Create Debian 11 (glibc 2.31) container suitable for cross-compiling Linux ppc64le binaries
```

platforms/linux-ppc64le/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+  curl https://sh.rustup.rs -sSf | sh -s -- -y \
+    --no-modify-path \
+    --profile minimal \
+    && \
+  rustup target add powerpc64le-unknown-linux-gnu && \
+  pip3 install meson
+
+# Handy for debugging the compiled targets in Highway (hwy_list_targets)
+#RUN apt-get install -y qemu-user-static
+#ENV QEMU_LD_PREFIX="/usr/s390x-linux-gnu"
```
kleisauke, line 41:
```suggestion
#ENV QEMU_LD_PREFIX="/usr/powerpc64le-linux-gnu"
```

platforms/linux-ppc64le/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+
+# Compiler settings
+ENV \
+  PKG_CONFIG="powerpc64le-linux-gnu-pkg-config --static" \
+  PLATFORM="linux-ppc64le" \
+  CHOST="powerpc64le-linux-gnu" \
+  RUST_TARGET="powerpc64le-unknown-linux-gnu" \
+  FLAGS="" \
+  WITHOUT_HIGHWAY="false" \
```
kleisauke, line 50: This still disables Highway, since this env variable is checked with `-z`.
https://github.com/lovell/sharp-libvips/blob/ad960df07ac72b911086a514683dbd346038ff3f/build/lin.sh#L320

```suggestion
```

sumitd2:
@kleisauke thanks for your input. I figured both of them out, plus another one. Working on the build failure, do you have any suggestions there?

kleisauke:
@sumitd2 I just opened PR #243 to fix these build failures.

sumitd2:
@lovell done

sumitd2:
@lovell @kleisauke darwin-x64 failed

sumitd2:
@lovell @kleisauke I am not sure how the npm/linux-ppc* directory should be named. js returns the arch as ppc64, not ppc64le.

lovell:
> I am not sure how the npm/linux-ppc* directory should be named. js returns the arch as ppc64, not ppc64le.

Good question. Node.js (and therefore npm) appears to use the endian-less `ppc64` for the [platform](https://nodejs.org/docs/latest/api/process.html#processarch) value.

What is the output of `node -p "process.arch"` on a ppc64le machine you're hoping to use?

sumitd2:
> node -p "process.arch"

[root@sumit-rhel ~]# node -p "process.arch"
ppc64

lovell (approved):

kleisauke (approved):

sumitd2:
@lovell @kleisauke The problem is here:
https://github.com/sumitd2/sharp-libvips/blob/8d2f34914f24cd36421e8b3b4baa8f08cfba9cf6/npm/populate.sh#L9

kleisauke:
@sumitd2 That line shouldn't cause issues for `linux-ppc64le`, it only does the following substitutions: 
`darwin-arm64v8` -> `darwin-arm64`
`linux-arm64v8` -> `linux-arm64`
`linux-armv6` -> `linux-arm`
`linuxmusl-arm64v8` -> `linuxmusl-arm64`

But indeed, we should probably also strip the `le` suffix there to match the endian-less npm package naming convention:
`linux-ppc64le` -> `linux-ppc64`

sumitd2:
> But indeed, we should probably also strip the `le` suffix there to match the endian-less npm package naming convention: `linux-ppc64le` -> `linux-ppc64`
@kleisauke Yes, that's what I meant. I am planning to check if the variable contains the string "ppc64le" and strip the last two characters, unless you have a better solution.

sumitd2:
@kleisauke I have updated populate.sh, can you trigger the build? Also, when do you plan publish it to the npm registry? My next task is to port lovell/sharp to Power, so I assume libvips will be needed in the registry.

lovell:
Thanks for the updates, I'm planning to rename some of the `platform` subdirectory names to match those under `npm` as a precursor to #206 anyway.

The next time this is published to npm will be as part of the next release of libvips, and yes, after that's done we'll need to do the other half of this task for prebuilt sharp binaries too.

I can't say when, but perhaps you could subscribe/watch for new releases.

lovell:
@sumitd2 The [@img/sharp-libvips-linux-ppc64](https://www.npmjs.com/package/@img/sharp-libvips-linux-ppc64) package is now published. If you're still interested, this means that the next step of adding prebuilt sharp binaries can now be considered, thank you.

PR #243, opened 2024-07-16
Bump deps: heif, rsvg

kleisauke:


lovell:
Brilliant, thanks Kleis.

PR #244, opened 2024-08-10
Workaround zune-jpeg issue

kleisauke:
See: https://gitlab.gnome.org/GNOME/librsvg/-/merge_requests/967#note_2088834


lovell:
Thanks Kleis, looking at the dependency chains involved here I agree this is a better workaround.

PR #249, opened 2024-10-03
Bump deps: archive, fribidi, glib, harfbuzz, png, rsvg, tiff, xml2, zlib-ng

kleisauke:


lovell:
Thanks Kleis, with 9 dependency updates this qualifies as a bumper bump.

PR #254, opened 2024-10-21
Bump dep: glib

kleisauke:
Additionally, change the libwebp release check back to https://release-monitoring.org/ to prevent the [webp-rfc9649](https://github.com/webmproject/libwebp/releases/tag/webp-rfc9649) tag from being identified as the latest version.


lovell:
Thanks Kleis, as always.

PR #255, opened 2024-10-21
Upgrade to libvips v8.16.0-rc2

kleisauke:


kleisauke:
We could perhaps backport https://github.com/strukturag/libheif/pull/1330.patch as a follow-up.

lovell:
I think there's a new version of libheif imminent - see https://github.com/strukturag/libheif/commit/f396bd73331fadb623604ada0a16f51c6ac5fb9b

Not sure why the s390x build is failing, best guess would be some kind of regression in serde (as a librsvg dependency).

The most recent passing build - https://github.com/lovell/sharp-libvips/actions/runs/11437971230
```
2024-10-21T10:24:42.8161684Z    Compiling serde v1.0.204
```
This build:
```
2024-10-22T07:23:36.5375908Z #6 66.61    Compiling serde v1.0.211
```

kleisauke:
Commit df811a1f596ead0e90f44b09b770b59222aa5b8a should resolve the failing build, which was impacting more than just s390x. It does indeed seem a regression in serde.

lovell:
Dankjewel!

kleisauke:
For reference, upstream serde issue: https://github.com/serde-rs/serde/issues/2844.

PR #259, opened 2024-12-11
Bump deps: glib, pango

kleisauke:


lovell:
Thanks Kleis, as expected those "ifunc" compilation errors are gone. :+1:

PR #260, opened 2024-12-12
pixman: ARM64 intrinsics supported on Mach-O

kleisauke:
Issue https://gitlab.freedesktop.org/pixman/pixman/-/issues/69 was resolved via https://gitlab.freedesktop.org/pixman/pixman/-/merge_requests/71.


lovell:
Thanks for tracking/updating this Kleis.

PR #261, opened 2024-12-12
rsvg: improve zune-jpeg workaround

kleisauke:


lovell:
Are the build failures due to multiple/conflicting versions of `zune-core` in the dependency tree?

> 2024-12-12T13:06:23.6334763Z    Compiling zune-core v0.4.11 (https://github.com/etemesi254/zune-image.git?rev=80e1957#80e19571)
> ...
> 2024-12-12T13:06:29.1246132Z    Compiling zune-core v0.4.12

kleisauke:
Indeed, that's weird. It looks like I didn't test it properly before submitting this.

wasm-vips takes a similar approach, which seems to work well, but that uses resvg instead.
https://github.com/kleisauke/wasm-vips/blob/3f143eca218043d53dcd17b765419731030c5075/build.sh#L450-L457

I'll abandon this PR and add a note to https://github.com/etemesi254/zune-image/pull/242.

kleisauke:
Re-opening now that zune-jpeg v0.4.14 was released.

PR #262, opened 2024-12-12
resvg: update licence

kleisauke:
See: https://github.com/linebender/resvg/commit/cfb2ee57d53852d26189adf0dc047b9b68b32c0c

As a side note, we might consider standardizing the spelling of license/licence across these files.


lovell:
I try to use "licensing" as that works everywhere ;)

The spelling should probably match that used in the licence itself (which is almost always going to be "license").

Given the dual-licensing of resvg, we should select the one that we are using it under; MIT is probably easiest.

kleisauke:
I wasn't sure how to mention dual-licensed libraries. We should likely do the same for Highway as follow-up, which was dual-licensed to ensure compatibility with NumPy's BSD-3 license (see e.g. the removed licensing paragraph in https://github.com/numpy/numpy/commit/f09ab71f7c52e182e4f6c947800a9d30d80cdc2a).

PR #264, opened 2025-02-19
Linux: ensure GLib's `g_print*()` functions wrap the system ones

kleisauke:
C99 `printf()` semantics are well-supported by these Linux cross-targets, avoids the need for gnulib compat functions.

An alternative is setting up QEMU when cross-compiling, as this functionality is gated by `meson.can_run_host_binaries()`, but this is a bit simpler.

Note: this is not needed for macOS due to:
https://github.com/GNOME/glib/blob/2.83.3/meson.build#L1204-L1212
(you can verify this by checking for the `GLIB_USING_SYSTEM_PRINTF` definition in `lib/glib-2.0/include/glibconfig.h`)

Windows binaries, where this was originally found, will be fixed via PR https://github.com/libvips/build-win64-mxe/pull/72 or work-in-progress commit https://github.com/libvips/build-win64-mxe/commit/ea91b64f9a56369b2abf2d2f224f91f2420f8337.

### DIFF
--- a/platforms/linux-armv6/meson.ini
+++ b/platforms/linux-armv6/meson.ini
@@ -13,6 +13,12 @@ ld = 'arm-rpi-linux-gnueabihf-gcc-ld'
 strip = 'arm-rpi-linux-gnueabihf-strip'
 ranlib = 'arm-rpi-linux-gnueabihf-gcc-ranlib'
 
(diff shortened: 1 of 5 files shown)


lovell:
:+1:

Issue #238, opened 2024-06-25
"Error: Could not load the "sharp" module using the win32-arm64 runtime"


wojtekmaj:
## Possible install-time or require-time problem

- [x] I have read and understood all of the [documentation relating to installation](https://sharp.pixelplumbing.com/install).
- [x] I have searched for known bugs relating to this problem in my choice of package manager.

You must confirm both of these before continuing.

### Are you using the latest version of sharp?

- [x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

If you cannot confirm this, please upgrade to the latest version and try again before opening an issue.

If you are using another package which depends on a version of `sharp` that is not the latest,
please open an issue against that package instead.

### Are you using a supported runtime?

- [x] I am using Node.js with a version that satisfies `^18.17.0 || ^20.3.0 || >=21.0.0`
- [ ] I am using Deno
- [ ] I am using Bun

If you cannot confirm any of these,
please upgrade to the latest version
and try again before opening an issue.

### Are you using a supported package manager and installing optional dependencies?

- [ ] I am using npm >= 9.6.5 with `--include=optional`
- [x] I am using yarn >= 3.2.0
- [ ] I am using pnpm >= 7.1.0 with `--no-optional=false`
- [ ] I am using Deno
- [ ] I am using Bun

If you cannot confirm any of these, please upgrade to the latest version of your chosen package manager
and ensure you are allowing the installation of optional or multi-platform dependencies before opening an issue.

### What is the complete error message, including the full stack trace?

```
node_modules\sharp\lib\sharp.js:114
  throw new Error(help.join('\n'));
        ^

Error: Could not load the "sharp" module using the win32-arm64 runtime
Possible solutions:
- Manually install libvips >= 8.15.2
- Add experimental WebAssembly-based dependencies:
    npm install --cpu=wasm32 sharp
    npm install @img/sharp-wasm32
- Consult the installation documentation:
    See https://sharp.pixelplumbing.com/install
    at Object.<anonymous> (node_modules\sharp\lib\sharp.js:114:9)
    at Module._compile (node:internal/modules/cjs/loader:1460:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1544:10)
    at Module.load (node:internal/modules/cjs/loader:1275:32)
    at Module._load (node:internal/modules/cjs/loader:1091:12)
    at wrapModuleLoad (node:internal/modules/cjs/loader:212:19)
[...]

PR #272, opened 2025-04-22
Publish Windows arm64 package to NPM

dennisameling:
Windows ARM64 binaries are already being built. For example, the most recent [8.16.1 release](https://github.com/lovell/sharp-libvips/releases/tag/v8.16.1) already contains a `libvips-8.16.1-win32-arm64v8.tar.gz` file.

Let's start publishing the `@img/sharp-libvips-win32-arm64` NPM package to unblock https://github.com/lovell/sharp/pull/4375.

Closes https://github.com/lovell/sharp-libvips/issues/238

### DIFF
--- a/npm/package.json
+++ b/npm/package.json
@@ -15,6 +15,7 @@
     "linux-s390x",
     "linux-x64",
     "win32-ia32",
-    "win32-x64"
+    "win32-x64",
+    "win32-arm64"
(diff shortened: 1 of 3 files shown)


lovell:
https://www.npmjs.com/package/@img/sharp-libvips-win32-arm64

PR #275, opened 2025-06-06
Upgrade to libvips v8.17.0

kleisauke:


lovell:
Thanks Kleis, I think I'm going to refactor out the version number checking logic from the build script (and merge the value of the libvips version into this) so it's quicker to check locally if there's a newer version of any dependency without running an entire build.

kleisauke:
Great, sounds good to me. 

I'm currently investigating this on the `darwin-x64` runner:
```
Checking for function "rsvg_handle_set_stylesheet" with dependency librsvg-2.0: NO 
```
https://github.com/lovell/sharp-libvips/actions/runs/15489943190/job/43612935538#step:4:9155

Curiously, this is not an issue on the `darwin-arm64v8` runner.
https://github.com/lovell/sharp-libvips/actions/runs/15489943190/job/43612935518#step:4:9010

(this was originally found in https://github.com/kleisauke/net-vips/commit/2253e0a8b494bf2b7720361938b465ed3bc76a31)

lovell:
Perhaps `prefix: '#include <librsvg/rsvg.h>'` is required here?

https://github.com/libvips/libvips/blob/5da295f49fe27a33a79100cd332fe401d6835264/meson.build#L392

kleisauke:
Good catch. I'll open a draft PR on this repository to see if it fixes the issue.

(sounds very similar to the issues linked to https://github.com/mesonbuild/meson/pull/11011)

lovell:
See commit https://github.com/lovell/sharp-libvips/commit/be68966878a788e4e78969552a2b338a0b1e42b9 for the CI improvements to version number checking (plus takes advantage of new ARM64 runners).

PR #227, opened 2024-03-24
Meson: uses system directories in machine files

kleisauke:
Noticed this while doing:
```console
$ mkdir linux-x64
$ curl -fsSL https://github.com/lovell/sharp-libvips/releases/download/v8.15.2/libvips-8.15.2-linux-x64.tar.gz | tar xzC linux-x64
$ strings linux-x64/lib/libvips-cpp.so.42 | grep "/fontconfig"
  <dir prefix="xdg">fonts</dir>  <cachedir>/target/var/cache/fontconfig</cachedir>  <cachedir prefix="xdg">fontconfig</cachedir>  <include ignore_missing="yes">/etc/fonts/conf.d</include>  <include ignore_missing="yes" prefix="xdg">fontconfig/conf.d</include>  <include ignore_missing="yes" prefix="xdg">fontconfig/fonts.conf</include></fontconfig>
/target/share/fontconfig/conf.avail
/target/var/cache/fontconfig
/fontconfig
```

With this PR, I see:
```console
$ strings libvips-8.15.2-linux-x64/lib/libvips-cpp.so.42 | grep "/fontconfig"
  <dir prefix="xdg">fonts</dir>  <cachedir>/var/cache/fontconfig</cachedir>  <cachedir prefix="xdg">fontconfig</cachedir>  <include ignore_missing="yes">/etc/fonts/conf.d</include>  <include ignore_missing="yes" prefix="xdg">fontconfig/conf.d</include>  <include ignore_missing="yes" prefix="xdg">fontconfig/fonts.conf</include></fontconfig>
/usr/share/fontconfig/conf.avail
/var/cache/fontconfig
/fontconfig
```

And this during configuring Fontconfig:
```diff
   Paths
-    Cache directory            : /target/var/cache/fontconfig
-    Template directory         : /target/share/fontconfig/conf.avail
+    Cache directory            : /var/cache/fontconfig
+    Template directory         : /usr/share/fontconfig/conf.avail
     Base config directory      : /etc/fonts
     Config directory           : /etc/fonts/conf.d
-    XML directory              : /target/share/xml/fontconfig
+    XML directory              : /usr/share/xml/fontconfig
```

PR #277, opened 2025-06-11
Ensure files are not placed outside install prefix

kleisauke:
Commit ec75c960c6a20bfc5bef134ddf181023c8bf1ddf and 5c27d097dcc77136178bce492a5f7e83b96a4998 ensures that files are not placed outside the install prefix. This allow us to restore the directories removed in commit 27607130f3c98d9694155107989be31654ebe6dd, which are possibly(?) required by Fontconfig to locate system fonts.

Note: upstream PR https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4661 would prevent the `/target/share/locale` build path from being included in the binaries (context: https://github.com/lovell/sharp-libvips/pull/227#issuecomment-2016828471).


lovell:
Thank you, I was rather hasty in my removal of the macOS cross-compilation config, and you're totally right about fontconfig needing these.

PR #279, opened 2025-07-02
Bump deps: fontconfig, png

kleisauke:
Also, try to revert commit f40ab0df45a30f26172d7ab1168e570c0fdbe52e (I suspect the issue was resolved in libpng 1.6.50) and revise commit df340ef602c4f109e0bc5f1eb15557053da47879 accordingly.


kleisauke:
Marking as draft since libpng 1.6.50 has not (yet) been uploaded to https://sourceforge.net/projects/libpng/files/libpng16/.

kleisauke:
Good idea! I've switched to using the tagged tarballs from GitHub instead, which is safe according to [`diffoscope`](https://diffoscope.org/) (i.e. identical extracted content when comparing v1.6.49, the uploaded tarballs on SourceForge were likely created using `git archive`).

lovell:
Briljant, dankjewel!

PR #283, opened 2025-07-08
linux-x64: use rust nightly

kleisauke:
Also, forcefully disable gold as the linker, as Rust still seems to emit a warning:
https://github.com/lovell/sharp-libvips/actions/runs/16145634431/job/45563658436#step:4:13810
(though, I think it's safe in that case, since we are building librsvg statically)

Note that we could use `export CARGO_PROFILE_RELEASE_TRIM_PATHS="all"` as a possible follow-up. However, enabling it isn't very straightforward, since `-Ztrim-paths` cannot be added to the `RUSTFLAGS` env variable, see:
https://doc.rust-lang.org/cargo/reference/unstable.html#profile-trim-paths-option

### DIFF
--- a/platforms/linux-arm64v8/Dockerfile
+++ b/platforms/linux-arm64v8/Dockerfile
@@ -44,6 +44,8 @@ RUN \
   ln -s /usr/bin/gcc10-strip /usr/bin/strip && \
   ln -s /usr/bin/gcc10-ranlib /usr/bin/ranlib && \
   ln -s /usr/bin/gcc10-readelf /usr/bin/readelf && \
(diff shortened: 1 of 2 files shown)


kleisauke:
Before:
```
-rw-r--r-- 1 root root 94095574 Jul  8 14:22 librsvg-2.a
-rwxr-xr-x 1 root root 16916968 Jul  8 14:23 libvips-cpp.so.8.17.1
```

After:
```
-rw-r--r-- 1 root root 92855122 Jul  8 15:18 librsvg-2.a
-rwxr-xr-x 1 root root 16642544 Jul  8 15:19 libvips-cpp.so.8.17.1
```

PR #287, opened 2025-08-30
Bump deps: glib, harfbuzz, pango

kleisauke:


lovell:
Dankjewel! Great work simplifying glib to remove the regex.

PR #296, opened 2025-09-25
Bump deps: expat tiff harfbuzz

woss:


woss:
wrong branch :)
i am trying to add packages so i can handle heic files. Basically doing what you are doing but publishing on my personal org. Sorry for false alarm

PR #314, opened 2026-01-28
libultrahdr: temporarily build from latest commit

kleisauke:
In preparation for migrating the Linux x64 and arm64 containers to Rocky Linux 8, which provides GCC 15 via the `gcc-toolset-15` package in the AppStream repository.


lovell:
Should we add https://github.com/google/libultrahdr/pull/383 also?

lovell:
Dankjewel Kleis!

PR #317, opened 2026-02-01
Linux: upgrade base images and therefore libc requirements

kleisauke:
- From Amazon Linux 2 to Rocky Linux 8 (glibc 2.28)
- From Alpine 3.15 to 3.20 (musl 1.2.5)


lovell:
Brilliant, thank you very much Kleis, I've merged these as separate commits.

PR #348, opened 2026-04-11
heif: build from commit for upstream fixes

lovell:
There are many useful changes upstream and I feel we're at the point where the risk of building from the latest commit is now lower than the risk of shipping an outdated version.

Also patches the RGBA encoder memory leak.

/cc @kleisauke


kleisauke:
Incorporated into the Windows builds and wasm-vips through the following two commits:
https://github.com/libvips/build-win64-mxe/commit/2d833bd1951be7c6e2937f3e15d59cf4064d9076
https://github.com/kleisauke/wasm-vips/commit/6f99463cba0fce4e92d1bc54924c9cd048aea572

(I've omitted the RGBA encoder memory leak patch, as it is no longer necessary after https://github.com/strukturag/libheif/commit/0af3fe269562e8712549ea97ebfe14692a4136fa, as confirmed by OSS-Fuzz.)

PR #135, opened 2022-05-28
Various improvements

kleisauke:
- Remove `CPATH` env in favor of `CPPFLAGS`.
- Remove redundant `LIBRARY_PATH` env.
- Remove redundant sed replacements and configure options.
- Enable Arm Neon support in libwebp, if available.
- Enable WebP support in libtiff (https://github.com/lovell/sharp/issues/3198).

Consolidated `linux-x64` build diff:
<details>
  <summary>Details</summary>
  
```diff
--- a/various-improvements-linux-x64.out
+++ b/various-improvements-linux-x64.out
@@ -4059,10 +4059,7 @@ checking whether to build static libraries... yes
 Configuring for a minimal library
 So far so good!
 Checking zlib
-checking zlib.h usability... yes
-checking zlib.h presence... yes
-checking for zlib.h... yes
-checking for gzread in -lz... yes
+checking for Z... yes
 Checking lzma
 Disabling lzma compression support
 Checking headers
@@ -8138,17 +8135,13 @@ checking for lerc_decode in -l"Lerc"... no
 checking Lerc_c_api.h usability... no
 checking Lerc_c_api.h presence... no
 checking for Lerc_c_api.h... no
-checking for WebPDecode in -lwebp... no
+checking for WebPDecode in -lwebp... yes
 checking webp/decode.h usability... yes
 checking webp/decode.h presence... yes
 checking for webp/decode.h... yes
 checking for X... no
 checking for the pthreads library -lpthreads... no
-checking whether pthreads work without any flags... no
-checking whether pthreads work with -Kthread... no
-checking whether pthreads work with -kthread... no
-checking for the pthreads library -llthread... no
-checking whether pthreads work with -pthread... yes
+checking whether pthreads work without any flags... yes
 checking for joinable pthread attribute... PTHREAD_CREATE_JOINABLE
 checking if more special flags are required for pthreads... no
 checking whether we are using the Microsoft C compiler... no
@@ -8211,7 +8204,7 @@ Libtiff is now configured for x86_64-pc-linux-gnu
 
   Installation directory:             /target
   Documentation directory:            ${prefix}/share/doc/tiff-4.4.0
-  C compiler:                         gcc -march=westmere -Os -fPIC -D_GLIBCXX_USE_CXX11_ABI=1 -fno-asynchronous-unwind-tables -ffunction-sections -fdata-sections -Wall -W
[...]

PR #354, opened 2026-04-23
tiff: switch from autotools to CMake, build from latest commit

lovell:
The `-DCMath_HAVE_LIBC_POW=1` flag is a bit of a hack to ensure libtiff's `HAVE_OPENGL` detection logic, which appears to always run despite setting `-Dtiff-opengl=OFF`, works when cross-compiling.


lovell:
@kleisauke you might be interested in making a similar change to `wasm-vips`

kleisauke:
Great! I'll make the same change into the Windows builds and wasm-vips tomorrow. It looks like the "Propagate -pthread into CFLAGS to ensure WebP support" comment can also be removed now (context: https://github.com/lovell/sharp-libvips/pull/135#discussion_r884167869).

kleisauke:
Incorporated into the Windows builds and wasm-vips through the following two commits:
https://github.com/libvips/build-win64-mxe/commit/31dae3567e622980b8b6d6d1602cddf97e1e40d4
https://github.com/kleisauke/wasm-vips/commit/dd5e6cd0e37e3692a3c950e276c3bc80dba87a02

I've also compiled some libraries explicitly with `-DCMAKE_BUILD_TYPE=MinSizeRel`, since using `-DCMAKE_BUILD_TYPE=Release` overrides the `CFLAGS=-Os` / `CXXFLAGS=-Os` environment variables (I learned that the hard way while investigating the increase in Wasm binary size).

PR #394, opened 2026-08-18
Add Linux LoongArch 64-bit support

cheungxi:
## Summary

  Add prebuilt libvips support for Linux LoongArch 64-bit (`linux-loong64`).

  ## Changes

  - Add an Ubuntu 24.04-based LoongArch cross-compilation environment.
  - Use the Ubuntu GCC 14 LoongArch cross-toolchain.
  - Add CMake and Meson cross-compilation configuration.
  - Configure the Rust `loongarch64-unknown-linux-gnu` target.
  - Add the `@img/sharp-libvips-linux-loong64` npm package.
  - Add `linux-loong64` to the build script, CI matrix, npm workspaces and documentation.
  - Enable Highway SIMD support using the LSX baseline with optional LASX runtime dispatch.
  - Declare a minimum glibc version of 2.39, matching the Ubuntu 24.04 build sysroot.

  ## Validation

  - Completed a full `linux-loong64` cross-build on an x86_64 Ubuntu host.
  - Successfully built libvips 8.18.5 and all enabled dependencies, including:
    - zlib-ng LoongArch optimisations
    - libffi LoongArch assembly
    - librsvg using the Rust LoongArch target
    - libheif and UltraHDR
  - Verified the resulting library is a LoongArch 64-bit ELF shared object.
  - Verified that only the expected system libraries remain dynamically linked:
    - `libstdc++.so.6`
    - `libm.so.6`
    - `libgcc_s.so.1`
    - `libc.so.6`
    - `ld-linux-loongarch-lp64d.so.1`
  - Tested the generated library on a native LoongArch system:
    - `vips_init()` returned `0`
    - `vips_version()` returned `8.18.5`
  - The `build-linux-loong64` GitHub Actions job completed successfully and uploaded its
  artifact.

  CI validation run:

  https://github.com/cheungxi/sharp-libvips/actions/runs/32037607798


lovell:
Hello, thank you for the PR.

As you've seen, this repo cross-compiles libvips binaries for various platforms and architectures. However the sharp repo uses QEMU-based emulation via [uraimo/run-on-arch-action](https://github.com/uraimo/run-on-arch-action) where native builds are unavailable. This does not currently support LoongArch CPUs so I am unable to merge and support this change at present.

cheungxi:
> Hello, thank you for the PR.
> 
> As you've seen, this repo cross-compiles libvips binaries for various platforms and architectures. However the sharp repo uses QEMU-based emulation via [uraimo/run-on-arch-action](https://github.com/uraimo/run-on-arch-action) where native builds are unavailable. This does not currently support LoongArch CPUs so I am unable to merge and support this change at present.

  Thank you. LoongArch64 support has now been merged and released in                         
  run-on-arch-action v3.2.0:                                                                 
                                                                                             
  https://github.com/uraimo/run-on-arch-action/releases/tag/v3.2.0                           
                                                                                             
  The action's LoongArch64 CI is passing:                                                    
                                                                                             
  https://github.com/uraimo/run-on-arch-action/actions/runs/32153980328                      
                                                                                             
  I will also prepare an end-to-end sharp QEMU build using a glibc                           
  linux/loong64 image. Could you please confirm whether the action release                   
  removes the platform-support blocker for this PR?

platforms/linux-loong64/Dockerfile
```diff
@@ -0,0 +1,65 @@
+FROM ubuntu:24.04
```
lovell, line 1: Can we use Debian here, perhaps 12 (bookworm, glibc 2.36) or 13 (trixie, glibc 2.41)?

lovell (commented):
Can we use one of the container images from https://github.com/loong64#docker-images for building binaries under QEMU in the sharp repo?

platforms/linux-loong64/Dockerfile
wojiushixiaobai: Refer to https://github.com/loong64/sharp-libvips/commit/105bafcd3a12d589bfdd7b98e435affcf91a17a8 for reference, 
but it uses an unofficial Debian repository, so it is recommended to keep the status quo (as-is).

- https://github.com/loong64/sharp-libvips/actions/runs/33131420064/job/98721595097

lovell:
I'd prefer that we use the same libc version for both the cross-compilation of sharp-libvips and the QEMU-emulated native compilation of sharp.

It looks like https://github.com/loong64/docker-library provides native Alpine and Debian, but cross-compiling on Debian seems more complex. Perhaps we should use Alpine instead? Is this a common choice for LoongArch users?

wojiushixiaobai:
@lovell 
To clarify - I meant to keep @cheungxi's changes as-is. Commit loong64@105bafc (run: https://github.com/loong64/sharp-libvips/actions/runs/33131420064/job/98721595097) is provided only as a proof of concept to show that this approach is possible, but it relies on an unofficial Debian repository and is not recommended.

